PR #138, opened 2020-01-27
Fix compilation issues

mfournial:
The issues came with update-alternatives being set to gcc, in particular
* comparison between signed `int` and `unsigned int` (`getNumCores`)
* C99 style initialization of a `struct` with `char*` in it (compiler issue)
* ignored return value from system calls

So this MR contains just a few GNU-recommended best practices now, as well as some fixes on the OpenMP cmake imports and some new runtime functions


.gitignore
```diff
@@ -3,6 +3,8 @@ perf/
 work/
 download/
 build/
+ninja-build/
+faasm.ini
```
Shillaker: `faasm.ini` shouldn't be in the root of the project, it should live at `~/faasm/faasm.ini`. I don't mind leaving it in `gitignore` but worth checking if you really need it there

func/omp/CMakeLists.txt
```diff
@@ -35,6 +35,7 @@ endfunction(omp_func)
 
 omp_func(checks checks.cpp false)
 omp_func(hellomp hellomp.cpp false)
+omp_func(create-parallel create-parallel.cpp false)
```
Shillaker: I've chosen convention of underscores in cmake targets, .h and .cpp files so let's stick with that rather than hyphens. If you see any I've missed it's worth changing them to underscores.

src/runner/GenericFunctionProfiler.cpp
```diff
@@ -18,6 +18,8 @@ namespace runner {
             throw std::runtime_error("Could not find binary at " + execPath);
         }
 
-        system(execPath.c_str());
+        int error = system(execPath.c_str());
+        if (error != 0)
```
Shillaker: Convention in this project is to put curly brackets with `if`s even if they're one liners

src/runner/PolybenchProfiler.cpp
```diff
@@ -15,6 +16,8 @@ namespace runner {
         std::string execPath = std::string(BINARY_DIR) + funcName;
 
         // TODO - is it possible to avoid system here?
-        system(execPath.c_str());
+        int error = system(execPath.c_str());
+        if (error != 0)
+            printf("Failed to execute %s", execPath.c_str());
```
Shillaker: Curly brackets

src/wasm/WasmModule.cpp
```diff
@@ -1020,10 +1021,16 @@ namespace wasm {
 
         // Make the fd big enough
         memoryFdSize = numBytes;
-        ftruncate(memoryFd, memoryFdSize);
+        int ferror = ftruncate(memoryFd, memoryFdSize);
+        if (ferror) {
+            logger->error("ferror call failed with error {}", ferror);
```
Shillaker: Error message and comments should start with caps.

src/wasm/syscalls_env.cpp
```diff
@@ -134,12 +134,12 @@ namespace wasm {
...
-                .release="1.0.0",
-                .version="Faasm 123",
-                .machine="x86",   // Probably safest in 32-bit wasm env
-                .domainname="(none)"
+                { .sysname="Linux" },
+                { .nodename="faasm" },
+                { .release="1.0.0" },
+                { .version="Faasm 123" },
+                { .machine="x86" },   // Probably safest in 32-bit wasm env
+                {.domainname="(none)" }
```
Shillaker: Was this really a compiler warning? I'm surprised as the other syntax doesn't seem that offensive to me.

src/wasm/syscalls_openmp.cpp
```diff
@@ -25,6 +25,19 @@ namespace wasm {
         return util::getUsableCores();
     }
 
+    WAVM_DEFINE_INTRINSIC_FUNCTION(env, "__kmpc_push_num_threads", void, __kmpc_push_num_threads, 
+                                   I32 loc, I32 global_tid, I32 num_threads) {
+        const std::shared_ptr<spdlog::logger> &logger = util::getLogger();
+        logger->debug("S - __kmpc_push_num_threads {} {}", global_tid, num_threads);
```
Shillaker: Include all arguments passed to the function in these logging calls (should be 3 here and 1 in __kmpc_global_thread_num)

src/wasm/syscalls_openmp.cpp
```diff
@@ -59,7 +72,7 @@ namespace wasm {
         Runtime::GCPointer<Runtime::Memory> &memoryPtr = getExecutingModule()->defaultMemory;
         
         // Spawn calls to the microtask in multiple threads
-        unsigned int numThreads = util::getUsableCores();
+        int numThreads = util::getUsableCores(); // or fetch local num_thread
```
Shillaker: Yeah I think we can probably avoid repeated calls to get the number of threads that are scattered all over.

func/omp/create-parallel.cpp
```diff
@@ -0,0 +1,22 @@
+#include <omp.h>
+#include <stdio.h>
+#include <faasm/faasm.h>
+
+void printThreadInfo(const char* label) {
```
Shillaker: When adding a new function to the codebase we should check in the `.wasm` file as well as the source. 

In this diff it looks like `hellomp.wasm` has changed but `create_parallel.wasm` hasn't been added.

Shillaker:
Looks good, just a few comments inline in the diff. 

Could you rebase with master to reduce the number of commits captured in this PR?

.gitignore
mfournial: I happened to check out faasm at `~/` on my VM!

src/wasm/syscalls_env.cpp
mfournial: I doubled checked: https://stackoverflow.com/questions/18731707/why-does-c11-not-support-designated-initializer-lists-as-c99

func/omp/CMakeLists.txt
```diff
@@ -1,7 +1,14 @@
 set(FAASM_USER omp)
 
-set(CMAKE_C_FLAGS "-fopenmp ${CMAKE_C_FLAGS}")
-set(CMAKE_CXX_FLAGS "-fopenmp ${CMAKE_CXX_FLAGS}")
+find_package(OpenMP QUIET)
```
Shillaker: It's a little confusing as to whether this is used for the wasm version (i'm assuming it always fails?). It would be clearer to have:

```
if(FAASM_BUILD_TYPE STREQUAL "wasm")
    set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -fopenmp")
    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fopenmp")
else()
    find_package(OpenMP QUIET)
    include_directories(${OPENMP_INCLUDE_DIR})
    set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} ${OpenMP_C_FLAGS}")
    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} ${OpenMP_CXX_FLAGS}")
endif()
```

ansible/local_dev.yml
```diff
@@ -4,6 +4,7 @@
   connection: local
   gather_facts: yes
   roles:
+    - eigen
```
Shillaker: `linux` role installs cmake so best to put this after it

docs/local_dev.md
```diff
@@ -4,7 +4,12 @@ This guide is only relevant for those wanting to dig deeper or make changes to F
 and run functions, then the Docker set-up described in the README should be sufficient. This includes deploying to 
 Knative or other cloud providers. 
 
-## Submodules
+## Checking out the repo
+
+Whether you are working remotely (see below) or locally, a lot of **default parameters** assume that the repository is
+located in `/usr/local/code/faasm`, the latter two directories _owned_ by the current user. To avoid having to set up
+environment variables in the many scripts you will use to get started, we recommend to **simlink** where you or your IDE
```
Shillaker: symlink*

docs/local_dev.md
```diff
@@ -116,3 +121,19 @@ To use cgroup isolation, you'll need to run:
 ```
 sudo ./bin/cgroup.sh
 ```
+
+## Remote development using CLion
+
+The first step you should be doing when CLion uploads the code to a temporary location (e.g. `/tmp/tmp.XXX`) is to
+simlink this location to `/usr/local/code/faasm` as explained above for consistency (e.g. VM crash).
```
Shillaker: symlink*

tasks/compile.py
```diff
@@ -36,21 +38,26 @@ def _do_compile(target, clean, debug):
...
     target = func
     _do_compile(target, clean, debug)
-
+    if cp:
+        cmd = [
+            "cp",
+            join(FUNC_BUILD_DIR, user, ".".join([func, "wasm"])),
+            join(WASM_DIR, user, func, "function.wasm"),
+        ]
+        call(" ".join(cmd), shell=True, cwd=FUNC_BUILD_DIR)
```
Shillaker: Nice

PR #144, opened 2020-01-29
Main OpenMP constructs

mfournial:
Few notes to start with on build related things:
* `rsync` the various `function.wasm` I create with OpenMP as discussed in PM
* Create tests that use those (e.g. test header API, simple_master, etc.)

Then this PR brings (for ** non teams implementation, assuming single master, no nested parallel**)
* [x] Simple `omp master` (non-nested, non-team) construct with a test
* [x] Simple `omp barrier` construct with a test
* [x] Fix a lot of the existing control behaviour (i.e. thread number, control from the user etc.) and includes tests to match native behaviour
* Delayed for future PR: `intel_nstreams.c`, from the Intel kernel repo, that with proper handling of `omp for schedule(static)` in a next PR will do. This will compile only natively, or if I add the OpenMP symbols will throw a runtime error or similar


func/omp/header_api_support.cpp
```diff
@@ -0,0 +1,38 @@
+#include <omp.h>
+#include <stdio.h>
+#include <faasm/faasm.h>
+
+bool failed = false;
+
+struct ThreadInfo {
+  int32_t thread_num;
+  int32_t local_thread_num;
+} __attribute__ ((aligned(64)));
```
Shillaker: Is the `aligned(64)` necessary? Not necessarily a problem, just interested

src/wasm/syscalls_openmp.cpp
```diff
@@ -1,40 +1,109 @@
...
+
 namespace wasm {
     static thread_local int thisThreadNumber = 0;
-    static thread_local unsigned int thisSectionThreadCount = 0;
+    static thread_local unsigned int thisSectionThreadCount = 1;
+    static int masterNumThread = -1;
+    static std::vector<std::thread> threads;
+    static int numThreadSemaphore = -1;
+    static std::mutex barrierMutex;
+    static std::condition_variable barrierCV;
```
Shillaker: Will this break if we have concurrent OpenMP functions running?

Shillaker: We need to support having multiple teams of threads executing different OpenMP functions as part of the same Faasm runtime process. I have a feeling they would interfere with each other as it is

func/omp/header_api_support.cpp
mfournial: Technically no, I just want to avoid false sharing when the threads are writing back to the same struct. Now that I think of it I think I meant `128`

mfournial: This will be moved to a global set of macros however like Clang does it

src/wasm/syscalls_openmp.cpp
mfournial: They absolutely would, the next step is to support teams, which this PR is not doing and do like clang which has a global `__kmp_threads` array indexable with the `gtid`. Supporting teams was not in the scope of this PR yet, as it will first require to fix the `gtid` problems I've seen.

Shillaker: Ok great, we also have to think about the multi-tenancy angle, i.e. totally isolating two potentially malicious users from each other within the same Faasm runtime. This will be a stricter form of isolation than between teams that are part of the same application. 

If the `gtid` is used to index resources but can be modified in wasm code, then we need to make sure each user has their own `gtid` space. In other words, a malicious user shouldn't be able to manipulate the `gtid` that gets passed to functions to try and access/ mess up another user's stuff.

Shillaker: But as you say, out of scope for this PR so will merge

mfournial: That notion of multi-user is something I'm looking at very closely, it is also why I preferred to re-implement the runtime rather than trying to get the current one to compile to WASM for example because I need to study how to integrate control of it.

There are other important questions here on top of the security concerns that you raised, things like fairness, enforceable SLAs, user-defined device & target programming.

PR #145, opened 2020-01-30
WIP: Add support for omp for schedule(static) with int param

mfournial:
In the same simple teamless setting of #144 without nesting, multiple constructs etc. adds support notably for `init_4` to now run programs of the kind 
```c
#pragma omp parallel for num_threads(2) schedule(static, 5) default(none) shared(results)
for (int i = 0; i < total; i++) {
    results[i].i = i * (1 + omp_get_thread_num());
}
```
This means the `intel_nstreams` file should now be working more or less as expected. More details will be added once the `simple_for` case is running. 

### TODOs

* [ ] Add `simple_for` to the tests
* [ ] Implement for `sched_type` 34 the required controls in `init_4`.
* [ ] Implement for `sched_type` 33 and/or `chunk` size 1 (task ticketing system like in Clang)
* [ ] Get `intel_nstreams` to compile and run on a single machine

PR #148, opened 2020-01-31
Add support for omp parallel for schedule(static, [chunk]) with int parameter

mfournial:
This contains the minimal error caused when using the `omp parallel for` construct.

The `simple_for.cpp` program makes the following calls to the runtime:
```
$  ltrace -l libomp.so.5 -f -s 128 -C -n 4 build/bin/simple_for
[pid 10620] simple_for->__kmpc_global_thread_num(0x7ffde77c4e60, 0x7fc3857ffd50, 80, 0x7ffde77c4e20)                                               = 0
[pid 10620] simple_for->__kmpc_push_num_threads(0x7ffde77c4e60, 0, 1, 0x48b530)                                                                    = 0x188d600
[pid 10620] simple_for->__kmpc_fork_call(0x7ffde77c4e60, 1, 0x40b6b0, 0x7ffde77c4e7c <unfinished ...>
[pid 10620]     simple_for->__kmpc_for_static_init_4(0x7ffde77c4a70, 0, 34, 0x7ffde77c4a90)                                                        = 0x7fc385a0bd4c
[pid 10620]     simple_for->__kmpc_for_static_fini(0x7ffde77c4a70, 0, 0x185a010, 0x7ffde77c4e7c)                                                   = 0x7fc385a0c59c
[pid 10620] <... __kmpc_fork_call resumed> )                                                                                                       = 0x7fc385a10328
[pid 10620] +++ exited (status 0) +++
```
A GDB watch on that variable to see when it changes would be very useful, is that something possible to do in Faasm?. Because where does this memory _corruption_ come from 🤔?


mfournial:
I was able to get a more firm understanding of things, I'll put the gist of things here because you might be interested.

Remember, the shared variables get erased upon reaching the `omp for` construct. It has to do with something in the lines of an ABI-like issue:
* Unfortunate incompatibility of some copy semantic for function parameters between C/C++/Wasm
* In the mix of OpenMP/WASM/JIT something somewhere is messing up the call stack of the OpenMP function created by the compiler (i.e. the loop body)
* This results in the original pointers to the **shared** data that the compiler had given us are **not valid anymore** 

Let me explain with an example. First, you need to understand that the compiler will create 4 variables for the static loop construct and give pointers to those to `__kmpc_for_static_init_4` which will modify their value for each thread:
* `lower`: LB on the array (it is however **not** necessarily the initialization value of `i`, your loop counter)
* `upper` inclusive UB on the array (like `lower`, the compiler shifts those if let's say you start looping at `i=3; i < 5`, in which case `lower` = 0, `upper` =1)
* `stride`: value for the runtime library to calculate
* `lastiter`: last iter flag.

Now that we know this let's create a stack array bigger than 4, say 15 elements, and keep the counting variable we had originally in `simple_for`.

```c++
FAASM_MAIN_FUNC() {
    int32_t stack[15] = {-1, -2, -3, -4, -5, -6, -7, -8, -9, -10, -11, -12, -13, -14, -15};
    int32_t var = BASE_VALUE + ITERATIONS;

    #pragma omp parallel for num_threads(1) default(none) shared(stack, var)
    for (int i = START; i < END; i++) {
        var--;
        (stack + 15)[i]=42; // Note +15 here, we write at the "moved/copied" location
    }

    printf("Printing some values before stack and some long after\n");
    for (int i = -3; i < 15 + 13; i++) {
        printf("%i, ", (stack)[i]);
    }
    printf("\nvar: %d\n", var);
    return 0;
}
```

```
12582848, 0, 100, -1, -2, -3, -4, -5, -6, -7, -8, -9, -10, -11, -12, 0, 2, 1, 0, -1, -2, 42, 42, -5, -6, -7, -8, -9, -10, -11, -12, 
```

The two first value are not interesting. The `100` is the (correct) value of `var`. Then we see `-1, ..., -12` which is the old `stack` array, `stack` points points at `-1`. Then we see `0, 1, 1, 0` corresponding to the `lastflag`, `stride`, `upper`, `lower` variables which **overrides** the end of the stack array (+1 I32 space, maybe the copied `var` which is nowhere to be found). Finally we have the move/copied `stack` array, modified in the correct places (42, 42) because of the offset `+15` we have applied (_note_ that adding the length of the array only works on an array of size 15, this is not bulletproof).

If you go back to what I suggest the causes of this may be. I think I need to first exclude `fork` to be the cause of this, maybe there is something wrong in my semantics of things. Or there is some ABI issues I need to understand.

func/omp/simple_for.cpp
```diff
@@ -0,0 +1,24 @@
...
+
+// Use base values else than 0 or 1 because they are often reset to that
+#define BASE_VALUE 100
+
+FAASM_MAIN_FUNC() {
+    int var = BASE_VALUE + ITERATIONS;
+
+    #pragma omp parallel for num_threads(1) default(none) shared(var)
+    for (int i = 0; i < ITERATIONS; i++) {
+        var--;
```
Shillaker: Will this result in a thread-safe operation? I don't think the C++ decrement is guaranteed to be atomic in all scenarios, let alone when it's been compile to machine code via wasm

mfournial: No it's not, but this is a sequential loop (setting `num_threads(1)` creates only 1 thread in fork that the main thread waits on in the `fork` call

mfournial:
I moved the various pieces brought in #145 in here as well. Which means this PR will replace it.

We will operate in the same simple team-less context of #144 without nesting, multiple constructs, users etc. I'm changing the title of this PR to reflect what has to be done. This will be the main "title" comment too.

# Add support for omp parallel for schedule(static, [chunk]) with int parameter

### TODOs

* [ ] Implement for `sched_type` 34 the required controls in `init_4`.
* [ ] Implement for `sched_type` 33 and/or `chunk` size 1 (task ticketing system like in Clang?)
* [ ] Add `simple_for` to the tests
* [ ] Add `for_static_schedule to the tests
* [ ] Fix race conditions currently present

src/wasm/syscalls_openmp.cpp
```diff
@@ -190,10 +208,120 @@ namespace wasm {
...
+     * increment and chunks for parallel loop and distribute constructs.
+     *
+     * See sched_type for supported scheduling.
+     *
+     * SAFETY: The implementation here is not thread safe at the moment. The pointers values, although set
+     * to the correct values for each thread right before calling this function (on native, COW, so this
+     * function's modifications to them should be thread-local but are not at the moment). This function will
+     * therefore only work if each thread's work between init and fini is mutually exclusive..
+     * TODO: Find a way to make this thread safe, the hard bit is that the compiler assumes that those
+     * variables, although at the same address, will be thread-local.
```
mfournial: @Shillaker Have you encountered something similar before? I can't see a way to do this without modifying the compiler or WAVM. I need those 4 variables to follow the same semantic as `thread_local`

src/wasm/syscalls_openmp.cpp
```diff
@@ -263,54 +273,58 @@ namespace wasm {
...
+
+        logger->debug("Thread {}: lower {}, upper {}, lastiter {}, stride {}", thisThreadNumber, *lower, *upper,
+                      *lastIter, *stride);
 
         if (sectionThreadCount == 1) {
-            *plastiter = true;
-            *pstride = (incr > 0) ? (*pupper - *plower + 1) : (-(*plower - *pupper + 1));
-            return;
+            throw std::runtime_error("Not yet supported running loop with one thread");
         }
```
mfournial: What evidence do you have of this? it works fine with me?

Shillaker: None actually, I was just trying to reduce the amount of code I needed to look at earlier. You can put that back now

PR #149, opened 2020-02-01
OpenMP For Experiment

Shillaker:
@mfournial I've started my own investigation here and created a version from scratch. This should result in a thread-safe operation (writing counts to a shared array). 

The output on my machine is

```
0 = 10
1 = 0
2 = 0
3 = 0
```

Suggesting only one thread is actually making the changes.

I'm not sure if it's relevant but I changed `omp_get_thread_num` to log the thread number it's returning, and it always seems to be zero in all the loops. 

I'll look into this more later or on Monday. Feel free to play around.


include/wasm/openmp_types.h
```diff
@@ -11,3 +11,6 @@ struct wasm_ident {
     I32 reserved_3;
     char const *psource;
 };
+
+// Types of source locations that call fork (defined in kmp.h)
+#define KMP_IDENT_WORK_LOOP 2
```
mfournial: This seem to be only used by OMPT (OpenMP tooling) in the original library, why is this necessary?

src/wasm/syscalls_openmp.cpp
```diff
@@ -123,22 +124,24 @@ namespace wasm {
...
+            throw std::runtime_error("Unsupported calling type for fork");
+        }
+
         // Retrieve the microstask function from the table
         Runtime::Object *funcObj = Runtime::getTableElement(getExecutingModule()->defaultTable, microtaskPtr);
         Runtime::Function *func = Runtime::asFunction(funcObj);
         IR::FunctionType funcType = Runtime::getFunctionType(func);
 
-        // Get reference to module memory
-        Runtime::GCPointer<Runtime::Memory> &memoryPtr = getExecutingModule()->defaultMemory;
```
mfournial: Why drop the`GCPointer`?

mfournial:
Note that the `simple_for` test you divided above partly tests the correctness of the scheduling of the threads. I was originally working on this and made similar tests now available in the rewritten PR [here](https://github.com/lsds/Faasm/pull/148/files?file-filters%5B%5D=.cpp&file-filters%5B%5D=.h&file-filters%5B%5D=.imports&file-filters%5B%5D=.wasm#diff-89b0bae12c30a331da508647788b2159) on file `for_static_schedule`.

The scheduling logic works fine, apart from the race condition I need to fix (see comment with your mention on #148). Now for the memory issues I opened this PR for originally, I'll chase the memory bugs if they actually appear in real programs. With heap allocated vectors, things should hopefully be fine for now. You can safely close this unless you want to look more into the memory issue example program I commented on #148 Yesterday.

include/wasm/openmp_types.h
Shillaker: I wanted to add some more checks in the fork call to make sure we weren't missing anything. I first wanted to check whether the source location information changed between `parallel for` and `parallel`, e.g. to indicate some different argument format. This flag seems to be defined alongside other potentially useful types of source location (https://github.com/llvm-mirror/openmp/blob/cb6f3c46a841e47718564ca40f49af8bcfc57c3a/runtime/src/kmp.h#L203) and I wanted to get an error if it ever changed. Turns out it doesn't, so this is probably useless.

src/wasm/syscalls_openmp.cpp
Shillaker: `GCPointer` helps WAVM to perform its own garbage collection. It's useful to track long-lived references to WAVM-related stuff, so that it'll warn us if we're not able to clean up the module once we're done with it. If we're just accessing a pointer in a transient manner (e.g. inside a short function like this), then there's no need to add the complexity.

PR #176, opened 2020-02-21
WIP: Add LULESH module

mfournial:
This is currently blocked by: 
* [x] OpenMP-only Wasm target compiles but crashes, will be fixed by #177 
* [ ] MPI-only Wasm target: see below
* [x] Ultimately adding a _mixed_ OMPI targets would be nice.
* [x] Native compilation handling in the faasm CMakefile instead of the LULESH one would be great for CI and overall integration, notably when compiling `all_X` targets to have native and Wasm targets be the same as much as possible to make testing and discovery as easy as possible.

### The MPI Wasm target

The MPI Wasm target would require some work to work if we wanted to get it working, it currently has two missing symbols:
* `MPI_REQUEST_NULL`
* `MPI_Waitall`

Defined in `openmpi/includes/mpi.h` as such:
```cpp
#define MPI_REQUEST_NULL OMPI_PREDEFINED_GLOBAL(MPI_Request, ompi_request_null)
OMPI_DECLSPEC extern struct ompi_predefined_request_t ompi_request_null;

OMPI_DECLSPEC  int MPI_Waitall(int count, MPI_Request array_of_requests[],
                               MPI_Status *array_of_statuses);
 ```

### CODE REVIEW

**Comment by mfournial** [comment]
The OMP version works :tada: but very very very slowly. it requires to decrease the OMP stack size massively. I tested the impact of changing a number of variables, I'm saving the results below

| number of levels | itx | nx | stack_size | native time | native CPU | faasm time | faasm CPU |
|------------------|-----|----|------------|-------------|------------|------------|-----------|
|         1        |  5  | 20 |     128    |     0.13    |     342    |    11.39   |    244    |
|         5        |  5  | 20 |     128    |     0.15    |     358    |    10.99   |    245    |
|        500       |  5  | 20 |     128    |     0.14    |     350    |    11.25   |    244    |
|         3        |  5  | 10 |     128    |     0.09    |     300    |    11.8    |    246    |
|         3        |  2  | 20 |     128    |     0.05    |     313    |    4.81    |    241    |
|         3        |  5  | 10 |     256    |     idem    |    idem    |  Exception |           |
|         3        |  2  | 10 |     256    |     idem    |    idem    |    2.51    |    237    |
|         3        |  10 | 20 |     32     |     0.28    |     378    |    23.57   |    246    |
|         3        |  20 | 20 |     32     |     0.56    |     385    |  Exception |           |

File: func/mpi/CMakeLists.txt [hunk 1]
```diff
@@ -3,15 +3,37 @@ set(FAASM_USER mpi)
...
+            ../../third-party/LULESH/lulesh-comm.cc
+            ../../third-party/LULESH/lulesh-init.cc
+            ../../third-party/LULESH/lulesh-util.cc
+            ../../third-party/LULESH/lulesh.cc
+            )
+    add_executable(${lulesh_exec} ${lulesh_files})
+    target_link_libraries(${lulesh_exec} faasm faasmpi dlmalloc)
+
+    #set(ALL_OMP_FUNCS ${ALL_OMP_FUNCS} ${lulush_exec} PARENT_SCOPE)
+endfunction(lulesh_mpi_target)
```
Shillaker: It looks like this is building lulesh files, hence should be added as a new target in the lulesh fork rather than in Faasm. It's also repeated over two files which would be nice to avoid.

I'd suggest having a look at the `compile_eigen` target to see how I'd recommend compiling a CMake project. It should be relatively straightforward to put this stuff into the Lulesh fork, then call `cmake` on it with the Faasm toolchain file. You can still include `target_link_libraries(xxx faasm faasmpi dlmalloc)` within the Lulesh CMake files, as these exist in the Faasm toolchain's sysroot

Shillaker:
@mfournial merging this now. Have commented out your experimental code so that you don't need to keep it elsewhere

PR #177, opened 2020-02-22
Add support for nested parallel regions

mfournial:
This is currently what is making #176 crash so it is an important OpenMP feature. Will use this to remove the `static` and follow the more proper notions of `gtid`. **Note** that this is not teams, as teams are reserved for multiple device contexts


mfournial:
This is now done and passes the tests. Note that I've also used this as an opportunity to refactor a lot of stuff that I promised I'd do at some point. All the static stuff has been moved to the `OMPLevel` class as it should be. And I improved the barrier slightly to make sure its behavior stays correct as a user keeps on reusing it in a loop (added overflow check that throws, will fix it if it actually becomes an issue)

Shillaker:
Nice one, LGTM

PR #178, opened 2020-02-23
Add run all user function feature to simple runner

mfournial:
Keeps previous cmd line behaviour intact, however makes the function argument optional which results in running all codegened functions. This is useful when I want to test I didn't break some omp functions, without having to set up the tests.

### Examples

```bash
# previously possible commands, cmd line API preserved
simple_runner mpi hellompi
simple_runner mpi hellompi 5

# new commands
simple_runner mpi # runs all codegened function in the MPI Wasm folder once
simple_runner mpi 5 # same but 5 times
```

### DIFF
--- a/src/runner/simple_runner.cpp
+++ b/src/runner/simple_runner.cpp
@@ -4,26 +4,73 @@
 #include <util/timing.h>
 #include <util/func.h>
 #include <zygote/ZygoteRegistry.h>
+#include <boost/filesystem/path.hpp>
+#include <boost/filesystem/operations.hpp>
 
+int runFunction(std::string &user, std::string &function, int runCount);
 
 int main(int argc, char *argv[]) {
     util::initLogging();
     const std::shared_ptr<spdlog::logger> logger = util::getLogger();
 
     int runCount;
-    if (argc == 3) {
+    std::string user;
+    std::string function = {};
+    if (argc == 2) {
         runCount = 1;
     }
+    else if (argc == 3) {
+        try {
+            runCount = std::stoi(argv[2]);
+        } catch (std::invalid_argument&) {
+            function =  argv[2];
+            runCount = 1;
+        }
+    }
     else if (argc == 4) {
+        function = argv[2];
         runCount = std::stoi(argv[3]);
     }
     else {
-        logger->error("Usage: simple_runner <user> <func> [run_count]");
+        logger->error("Usage: simple_runner <user> [func] [run_count]");
         return 1;
     }
 
-    std::string user = argv[1];
-    std::string function = argv[2];
+    user = argv[1];
+
+    if (function.empty()) {
+        util::SystemConfig &conf = util::getSystemConfig();
+        logger->info("Running codegen for user {} on dir {} with {} runs", user, conf.functionDir, runCount);
+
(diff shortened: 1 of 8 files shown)


src/runner/simple_runner.cpp
```diff
@@ -4,26 +4,73 @@
...
         return 1;
     }
 
-    std::string user = argv[1];
-    std::string function = argv[2];
+    user = argv[1];
+
+    if (function.empty()) {
+        util::SystemConfig &conf = util::getSystemConfig();
+        logger->info("Running codegen for user {} on dir {} with {} runs", user, conf.functionDir, runCount);
```
Shillaker: Copy-paste error (not running codegen)

src/runner/simple_runner.cpp
```diff
@@ -55,12 +104,15 @@ int main(int argc, char *argv[]) {
         PROF_END(execution)
 
         if (result != 0) {
-            throw std::runtime_error(fmt::format("Non-zero return code ({})", result));
+            logger->error("Non-zero return code ({})", result);
+            res = 0;
```
Shillaker: `res = 1` or `res=result`?

tasks/compile.py
```diff
@@ -99,3 +101,22 @@ def _ts_compile(func, optimize=True):
...
     call(cmd_string, cwd=TS_DIR, shell=True)
+
+def _codegen_user(user, debug):
+    cmd = " ".join([
+        "codegen_func", # Requires to have bin dir in PATH
+        user,
+    ])
+    if debug:
+        print("Calling {}".format(cmd))
+    call(cmd, shell=True)
```
Shillaker: This is duplicating functionality that exists in `codegen_for_user`, however, that doesn't expect `codegen_func` to be on the `PATH` as this does (which is a good idea). It currently needs to run in Dockerfiles and set-up scripts, so checks certain dirs for the codegen binaries. Perhaps it could be changed to fall back to the path (i.e. modifying `find_codegen_xxx` to use the path as the first/last resort).

`codegen_for_user` is also a bit long-winded, I prefer `codegen-user`

tasks/compile.py
```diff
@@ -79,6 +78,9 @@ def compile_user(ctx, user, clean=False, debug=False, cp=False):
                 join(dest_folder, "function.wasm"),
             ]
             call(" ".join(cmd), shell=True, cwd=FUNC_BUILD_DIR)
+        if run:
+            _codegen_user(user, debug)
+            _simple_run_user(user, debug)
```
Shillaker: I'd rather keep tasks single-purpose and leave it up to the user to chain them if they want (`inv` allows you to chain multiple tasks in a single command). 

Simple runner isn't yet encapsulated in Python though, which thus makes that suggestion impossible. That would be nice and would have same syntax, i.e. `inv run <user> [func]`. This flow would then look like:

```
inv compile-user <my_user> codegen-user <my_user> run <my_user>
```

Shillaker: @mfournial this is now possible, with a slight alteration. I've created `run-user` which just takes a user argument, and `run` which takes a function too (i.e. `run-user mpi` and `run mpi hellompi`). This is to avoid having to do something like `run mpi` and `run mpi --func=hellompi`)

PR #188, opened 2020-03-08
Rust support

mfournial:
> Outdated message

This is only for reference, to show the print hack I'm doing at the moment. There is no real good way to do it otherwise at the moment

See the FFI Rust bindings crate: https://github.com/mfournial/faasm-sys  
and the wrapper crate: https://github.com/mfournial/rust-faasm


mfournial:
I tried to respect the conventions you set and make something that still worked with my [faasm-sys](https://github.com/mfournial/faasm-sys). I'll extend `rust-faasm` to fully support the host interface API elegantly and a bit more efficiently (e.g. I'm allocating for `println` which is unnecessary). I think I'll try to see if I can go lower lever like you did with println to not have to write glue code over and over again, but that's for later, for now I'll need a review please :)

Shillaker:
Awesome, looks great on first glance. I should be able to take a look later today

bin/download_libs.sh
```diff
@@ -0,0 +1,31 @@
+#!/bin/bash
+
+set -e
```
Shillaker: @mfournial I haven't actually used this but I think it might be useful when building the `faasm-sys` crate. It just downloads the latest `sysroot` and `toolchain`, thus avoiding the need to have a full dev environment set up just to build that crate. Does that make sense or does it require more than that?

mfournial: I understand you took things one step further than me, by trying to support both dev/normal install types of installations. I hard-coded the dev-library location in the build-script: https://github.com/mfournial/faasm-sys/blob/master/build.rs#L14 because I am waiting for the best way to pull Faasm system libraries (if they are not installed system-wide).

So I think this is a bit of an ad-hoc installation method for Faasm, so much that I wonder if this belongs to this repo or `faasm-sys`. I think supporting only dev is fine for Rust for now because things are unfinished and likely to change on both Faasm and Rust sides. But if you wanted to make this work I'd see it like this with the two install modes that faasm sys can switch between with a flag given to `cargo`:
* **dev mode**: Consider Faasm installed as `local_dev` and use the hardcoded path in the buildscript. Optional step: dynamically generate the Rust bindings to the library in the build script (currently [buggy](https://github.com/mfournial/faasm-sys/blob/master/build.rs#L33)) if there was any change on `lib-cpp` or `lib-rust`. Build rest of the crate normally. `TODO`: update build script to find and link the non Wasm libs to be able to run natively too.
* **Install mode/using faasm-sys as dependency e.g. for `rust-faasm`**: Check if Faasm wasm libs are installed in the build script if not download/update them from there. Same for native librs, normal dep from there

Shillaker: I think I'll get rid of this from the main repo as it now exists (as you suggest) in `faasm-sys`.

I agree re. dev and non-dev modes, and Rust being dev-only for now. That said, the non-dev stuff can be useful for CI and building containers (as you mentioned in the `faasm-sys` PR).

mfournial: I'm currently implementing both modes in Faasm-sys for CI

PR #217, opened 2020-04-16
Omp multihost simple reduce

mfournial:


mfournial:
Just a few notes on how I think distributed experiments could be simplified

When trying to create new distributed experiments, the code encourages more to design around previous experiments (e.g. MPI, TS, thread support, now OpenMP, etc.) and extend existing code. I see two problems in this in that new experiments get trickier to implement and they might break existing ones (issue notably in that distributed experiments are not in the CI).

I'm thinking of two refactors that would help:
1. Simplify the State KeyValue API (less custom types, more control). I don't think you intend to swap out Redis more than you write new experiments.
2. Improve execution dispatch mode, notably by simplifying the message protobuff (Too many bools if you see what I mean ;), also using fixing function arguments)

tasks/util/invoke.py
```diff
@@ -115,6 +117,11 @@ def invoke_impl(user, func,
     else:
         headers = {}
 
+    # deletes warm set
+    if knative and reset:
+        r = redis.Redis(host='localhost', port=6379, db=0)
+        r.delete("w_{}/{}".format(user, func))
```
Shillaker, line 123: This doesn't work in a distributed context, which is the primary mode of operation, hence probably shouldn't live here. There are separate tasks for managing Redis that work in all contexts, have a look at `redis.clear_queue`

src/wasm/openmp.cpp
```diff
@@ -31,59 +35,6 @@ namespace wasm {
...
-
-        // Returns user preference if set or device's maximum
-        return nextWanted > 0 ? nextWanted : util::getUsableCores();
-    }
-
-    // Thread-local variables for each OMP thread
-    thread_local int thisThreadNumber = 0;
-
     // Arguments passed to spawned threads. Shared at by the level apart from tid
     struct OMPThreadArgs {
```
Shillaker, line 39: Is this struct still needed? If not could you remove it?

include/state/StateBackend.h
```diff
@@ -65,6 +65,24 @@ namespace state {
          * Dequeues multiple values from a queue
          */
         virtual void dequeueMultiple(const std::string &queueName, uint8_t *buff, long buffLen, long nElems) = 0;
+
+        /**
+         * Increments the number stored at key by increment. If the key does not exist, it is set to 0 before
+         * performing the operation. This operation is limited to 64 bit signed integers.
+         * Returns new value.
+         */
+         virtual long incrByLong(const std::string &key, long value) = 0;
```
Shillaker, line 74: Would be nice to a see a test for these new functions

src/knative/worker_knative.cpp
```diff
@@ -16,11 +52,11 @@ int main() {
 
     // Start endpoint (will also have multiple threads)
     logger->info("Starting knative endpoint");
```
Shillaker, line 54: Useful to log the port here

src/openmp/Level.cpp
```diff
@@ -0,0 +1,45 @@
+#include "openmp/Level.h"
+
+namespace wasm {
```
Shillaker, line 3: As `openmp` is now a top-level `src` directory I think the namespace should just be `openmp` rather than `wasm::openmp`

src/wasm/openmp.cpp
```diff
@@ -292,63 +245,113 @@ namespace wasm {
...
+            return;
+        }
+
         // Set up number of threads for next level
-        int nextNumThreads = get_next_level_num_threads(thisLevel);
+        int nextNumThreads = thisLevel->get_next_level_num_threads();
         thisLevel->pushed_num_threads = -1; // Resets for next push
 
         // Set up new level
         OMPLevel *nextLevel = new OMPLevel(thisLevel, nextNumThreads);
```
Shillaker, line 350: `auto nextLevel = ...`

src/wasm/openmp.cpp
```diff
@@ -371,13 +374,16 @@ namespace wasm {
             }
 
             // Arguments for spawning the thread
-            // NOTE - CLion auto-format insists on this layout...
+            // NOTE - CLion auto-format insists on this layout... and clangd really hates C99 extensions
```
Shillaker, line 377: Yeah this is annoying

src/wasm/openmp.cpp
```diff
@@ -581,11 +587,18 @@ namespace wasm {
      */
     WAVM_DEFINE_INTRINSIC_FUNCTION(env, "__kmpc_reduce", I32, __kmpc_reduce, I32 loc, I32 gtid, I32 num_vars,
                                    I32 reduce_size, I32 reduce_data, I32 reduce_func, I32 lck) {
-
+        throw std::runtime_error("RIP");
```
Shillaker, line 590: Worth logging before throwing the exception to know what's killed it (same applies to other dead methods in here)

src/knative/CMakeLists.txt
```diff
@@ -2,6 +2,11 @@ include_directories(
         ${FAASM_INCLUDE_DIR}/knative
 )
 
+FIND_PACKAGE(Boost 1.65 QUIET OPTIONAL_COMPONENTS program_options)
+if (Boost_FOUND)
+    INCLUDE_DIRECTORIES(${Boost_INCLUDE_DIR})
+endif ()
+
```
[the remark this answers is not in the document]
Shillaker: Moved in latest commit

src/openmp/Level.cpp
mfournial: it doesn't need to be top level anymore, I am moving this down to a subdirectory in was. I originally forced this to be a nested namespace to remind me to change it later, but since you seem to be tying namespaces to directory hierarchy I'll follow that and move it

mfournial:
> I agree that the experiment code is not written to be extended, trying to create a one-size-fits-all experiment framework is a fool's errand, it's better to write each one from scratch in my experience. The stuff that's in there now should be moved into a separate repo as it's not really meant for public consumption (that said, experiments included in the current paper draft need to be viable for at least a few months from now). I'm reluctant to include experiments in CI as by definition they're somewhat ephemeral, and should be tied to a specific version of the code.

I agree with all of that.

>  I'm currently working on distributed in-memory state which would remove Redis
If you're doing that then don't touch anything indeed. It's just that there is 4 levels of abstractions to reach Redis, so just drop this comment.

> "execution dispatch mode" and what specifically is wrong with the protobuf object. 

There is a single WasmModule `execute` function which is going to keep on growing as the number of experiments grows and the protobuf obj becomes bloated. Since the person creating the protobuff obj always known in the first place which branch in the execute function it wants to execute it might be worth splitting this function along with the various kinds of object (i.e. execute python for python kind of message, execute OpenMP for OpenMP message etc.)

src/wasm/openmp.cpp
mfournial: Woops those are still meant to run if not distributed!
I need a way to elegantly handle the "am I in a distributed Level" paradigm. I was planning to do this in a later MR by extending the `OMPLevel` class into two subclasses and re-implement those functions in that. Because that would bloat that MR

src/knative/CMakeLists.txt
```diff
@@ -15,4 +20,9 @@ target_link_libraries(worker_knative_lib endpoint worker_lib)
 
 add_executable(worker_knative worker_knative.cpp)
 
-target_link_libraries(worker_knative worker_knative_lib)
+if (Boost_FOUND)
+    target_compile_definitions(worker_knative PRIVATE WITH_BOOST=1)
+    target_link_libraries(worker_knative worker_knative_lib Boost::program_options)
+else ()
+    target_link_libraries(worker_knative worker_knative_lib)
+endif ()
```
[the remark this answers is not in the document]
mfournial: boost filesystem is present by default on many distributions. not `program_options`

include/openmp/ThreadState.h
```diff
@@ -0,0 +1,18 @@
...
+
+namespace wasm {
+    namespace openmp {
+        // TLS for each OMP thread
+        extern thread_local int thisThreadNumber;
+        extern thread_local OMPLevel *thisLevel;
+
+        void setThreadNumber(int);
+
+        void setThreadLevel(OMPLevel *);
```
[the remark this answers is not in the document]
mfournial: The lifetime of it is well-defined so ARC is really overkill. I spent a lot of time trying to handle this neatly, as there are issues with initialisation, TLS, etc. I agree though that leaving a pointer to a shared object is never a good idea in the long run, but in the grand scheme of things it is premature to fix this now, as the handling of the OMPLevel is likely to change a bit more in the future

src/knative/worker_knative.cpp
```diff
@@ -1,11 +1,47 @@
...
+        if (vm.count("help")) {
+            std::cout << desc << std::endl;
+            return EXIT_SUCCESS;
+        }
+
+        port = vm["port"].as<int>();
+    }
+    catch(std::exception& e) {
+        std::cerr << e.what() << std::endl;
+        return EXIT_FAILURE;
```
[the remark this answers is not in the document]
mfournial: Yeah I'll remove things here. although with the number of cli utilities there is in the project, I think it's useful to have good support for command line args by default. I wanted to refactor `simple_runner` too for example, but one thing at a time on this PR, otherwise it's going to remain open forever

src/openmp/Level.cpp
Shillaker: :+1:

Shillaker:
Nice one, this is looking great and thanks for tidying up some of my historical laziness with that config, very much appreciated. Looks like Actions has recovered so hopefully the tests will run through and we can get this merged.

include/openmp/ThreadState.h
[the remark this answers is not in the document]
mfournial: What I feared would happen actually happened and when you refactored with the prepare OMP function you leaked and lead to tons of funny things.

I changed to `shared_ptr` because there is no way by manual memory management to get this right without using the stack and the const instantiation with the ternary operator hack that I did before but it pains me because we're paying that price simply because of an initial instantiation issue. I'll get back to this definitely, it'll be a lot cleaner after we do the few refactors I talked about below

include/openmp/Level.h
```diff
@@ -0,0 +1,47 @@
...
+
+namespace wasm {
+    namespace openmp {
+
+        // Global variables controlled by level master
+        class OMPLevel {
+        public:
+            const int depth = 0; // Number of nested OpenMP constructs, 0 for serial code
+            const int effective_depth = 0; // Number of parallel regions (> 1 thread) above this level
+            int max_active_level = 1; // Max number of effective parallel regions allowed from the top
```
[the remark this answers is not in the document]
mfournial: Will do this in a later PR since this class will need to change anyway. It's a bit confusing to be in the middle of Clang OpenMP which (righly ;)) uses snake case and Faasm, but I agree it clashes too much so I'll change this and also some in `openmp.cpp`

include/wasm/openmp/Level.h
```diff
@@ -0,0 +1,43 @@
...
+
+        // Global variables controlled by level master
+        class OMPLevel {
+        public:
+            const int depth = 0; // Number of nested OpenMP constructs, 0 for serial code
+            const int effective_depth = 0; // Number of parallel regions (> 1 thread) above this level
+            int max_active_level = 1; // Max number of effective parallel regions allowed from the top
+            const int num_threads = 1; // Number of threads of this level
+            int wanted_num_threads = -1; // Desired number of thread set by omp_set_num_threads for all future levels
+            int pushed_num_threads = -1; // Num threads pushed by compiler, valid for one parallel section, overrides wanted
```
Shillaker, line 19: Camel case

include/wasm/openmp/ThreadState.h
```diff
@@ -0,0 +1,18 @@
+#ifndef FAASM_THREADSTATE_H
+#define FAASM_THREADSTATE_H
```
Shillaker, line 2: `#pragma once`

src/knative/KnativeEndpoint.cpp
```diff
@@ -2,10 +2,15 @@
 #include "KnativeEndpoint.h"
 
 namespace knative {
-    KnativeEndpoint::KnativeEndpoint() : Endpoint(8080, 4) {
+
+    KnativeEndpoint::KnativeEndpoint() : Endpoint() {
+    }
+
+    KnativeEndpoint::KnativeEndpoint(int port, int threadCount) : Endpoint(port, threadCount) {
```
Shillaker, line 9: Do we still need this constructor?

include/endpoint/Endpoint.h
```diff
@@ -1,16 +1,19 @@
...
+
         Endpoint(int port, int threadCount);
         void start();
 
         virtual std::shared_ptr<Pistache::Http::Handler> getHandler() = 0;
     private:
-        int port;
-        int threadCount;
+        int port = util::getSystemConfig().endpointPort;
+        int threadCount = util::getSystemConfig().endpointNumThreads;
```
Shillaker, line 17: Can these be set in an initialiser list on the constructor implementation rather than in this header?

mfournial: No because there is more than one constructor: https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rc-in-class-initializer

You also always want a constructor in which you can pass them in for testing, same for the subclasses. 
There is a very good CPPconf (2018?) video about how this is this too that agrees with Bjarne Stroustrup on this.

Shillaker: Is that link not talking about constants though? I'm also not sure that having more than one constructor means you shouldn't use initialiser lists either. These are variables that will be changed by different system configurations and different constructors.

For checking changes in the system config in tests, we modify the SystemConfig singleton or corresponding environment variable (arguably replicating the real behaviour of the system more, but breaking from a strict unit test). Therefore I'm not sure I buy the testing angle.

I'm not that bothered by it, it just seems out of sync with the constructor and testing style in the rest of the codebase (which of course may be wrong). I'm inclined to stick with consistency and make wholesale changes if we want to change how we do something rather than have things out of sync.

PR #224, opened 2020-04-22
Speed comparison experiment

mfournial:


mfournial:
@Shillaker Obviously I'll remove everything in here that is in `docs`, I'm just using this as a way for you and Peter to be able to see this interesting result.

What about the experiment code, do you want it? if so where do I put it (I put it in `demo` to not be linked to OpenMP etc. but not sure it belongs there)

tasks/compile.py
```diff
@@ -15,8 +15,7 @@
 def _copy_built_function(user, func):
     dest_folder = join(WASM_DIR, user, func)
 
-    if not exists(dest_folder):
-        mkdir(dest_folder)
+    makedirs(dest_folder, exist_ok=True)
```
Shillaker, line 18: Nice, didn't know that existed. There a few places I've done the `if not exists...` thing so I'll keep an eye out :eyes:

PR #226, opened 2020-04-23
Add LLVM OpenMP project to Faasm toolchain

mfournial:
This fixes the OpenMP functions not compiling on the Faasm toolchain.

This PR contains two more things though:
* Uses Clang 10 (i.e. `/usr/bin/clang`) to compile the toolchain instead of a possibly old `cc`.
* I had to add the `/usr/local/include` import to be able to run `inv libs.faasm`. Possibly a bug? because this is definitely a standard path. So I don't really know...

### DIFF
--- a/libs/cpp/CMakeLists.txt
+++ b/libs/cpp/CMakeLists.txt
@@ -6,7 +6,7 @@ project(libfaasm)
 set(CMAKE_CXX_STANDARD 17)
 
 set(PROJ_INCLUDE ${CMAKE_CURRENT_LIST_DIR}/../../include)
-include_directories(${CMAKE_CURRENT_LIST_DIR} ${PROJ_INCLUDE})
+include_directories(${CMAKE_CURRENT_LIST_DIR} ${PROJ_INCLUDE} /usr/local/include)
 
 set(PUBLIC_HEADERS
         ${PROJ_INCLUDE}/faasm/core.h
--- a/libs/cpp/CMakeLists.txt
+++ b/libs/cpp/CMakeLists.txt
@@ -6,7 +6,7 @@ project(libfaasm)
 set(CMAKE_CXX_STANDARD 17)
 
 set(PROJ_INCLUDE ${CMAKE_CURRENT_LIST_DIR}/../../include)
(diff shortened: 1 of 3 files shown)


mfournial:
I compiled the toolchain and tested it by the way, this now compiles OpenMP functions

mfournial:
@Shillaker This now also includes a fix to the `randomInteger` test that this PR somehow managed to get unlucky enough to trigger!

libs/cpp/CMakeLists.txt
```diff
@@ -6,7 +6,7 @@ project(libfaasm)
 set(CMAKE_CXX_STANDARD 17)
 
 set(PROJ_INCLUDE ${CMAKE_CURRENT_LIST_DIR}/../../include)
-include_directories(${CMAKE_CURRENT_LIST_DIR} ${PROJ_INCLUDE})
+include_directories(${CMAKE_CURRENT_LIST_DIR} ${PROJ_INCLUDE} /usr/local/include)
```
Shillaker, line 9: I don't feel good about this, it has the potential to introduce some gnarly issues depending on the build environment. What's the error caused without this?

mfournial: Those lines in `matrix.h`:
```cpp
#include <eigen3/Eigen/Core>
#include <eigen3/Eigen/SparseCore>
```
I agree it shouldn't be in there and should be deleted in the future, I'm happy with having it only locally instead

Shillaker: Have you run `inv libs.eigen`? If you run `inv libs.toolchain` it should rebuild everything you need for a new toolchain. Think this is in the toolchain docs.

mfournial: Oh I remember now running into this before and needing to run `inv libs.eigen` thanks that fixes it :+1:.

I want to add a warning for me for `inv libs.faasm` because I treat it too often like it is `inv libs.toolchain`!

Shillaker:
Nice, LGTM. I'll run through the toolchain build and create a new release.

PR #229, opened 2020-04-23
Remove LLVM8 and Update to LLVM9

mfournial:
* Switch to LLVM detection by `llvm-config` instead of `LLVM_DIR` to give choice of LLVM version to user without needing to modify the `CMakeFile.txt` config (and will be consistent with `/usr/bin/X` compiler options for clarity and simlinks on llvm install). This can also now be overridden by a parameter anyway. (I check and this doesn't break building the docker containers too if `cpp-root` is rebuilt.)
* Update the `cpp-root` container to LLVM-9 (for good measure, not essential).
* Add missing simlinks to `cpp-root` container, notably `llvm-config` simlink for point above.
* Delete LLVM8 playbook
* Fix issues with LLVM 9 and 10 playbooks


tasks/docker.py
```diff
@@ -11,6 +11,7 @@
 
 # Order matters here
 RELEASE_CONTAINERS = [
+    "cpp-root",
```
Shillaker, line 14: Does this need to be here? This is the containers that get rebuilt for every release. `cpp-root` just contains dependencies and no application code. Additionally, the Docker release build is heavyweight enough as it is, I'd rather not introduce more steps.

CMakeLists.txt
```diff
@@ -78,8 +78,8 @@ if (${FAASM_CUSTOM_LLVM})
     set(WAVM_PERF_LIBS 1)
     add_definitions(-DWAVM_PERF_EVENTS=1)
 else ()
-    # Use LLVM 8 explicitly for the lib parts (i.e. codegen)
-    set(LLVM_DIR /usr/lib/llvm-8)
+    # Use LLVM 9 explicitly for the lib parts (i.e. codegen)
+    set(LLVM_DIR /usr/lib/llvm-9)
```
Shillaker, line 82: Can we not just go straight to 10? https://github.com/WAVM/WAVM/blob/master/CMakeLists.txt#L108

tasks/docker.py
mfournial: I agree that it's a bit pointless to push it, especially if you have it on your system already. My intention was more to make sure the `cpp-root` container would get built next time you run the release command. This `release` command is reserved for you anyway so I agree I shouldn't touch it.

I will instead make a separate "build all containers" command. Which will allow me for example to test changes to them or allow new-commers to reliably test them.

As for the speed issue since it seems that it is still not possible to have parallel builds in Docker I propose to you to move it to Github Actions that will be triggered on tagged builds (i.e. only on master)? Either running the normal runner, or making a self-hosted runner on one of the Koalas machine?

CMakeLists.txt
mfournial: That's a typo, this line was deleted as per the first bullet point in the PR description I don't know what happened to that commit!

mfournial: Fixed in https://github.com/lsds/Faasm/pull/229/commits/0db948cb63fa1221aa13cb23f1c042024a6261bd

tasks/docker.py
Shillaker: `cpp-root` is purely used to capture bits of the `base` container that don't often get rebuilt, but are expensive to build when they do. As a result, I'm happy to leave its rebuilding to be ad-hoc for now and just focus on the stuff that really needs to get rebuilt (i.e. the release containers). I'm not sure a "build all containers" command is very useful, and would inevitably not get maintained (i.e. new containers would accidentally get left off the list). Even when developing `cpp-root` surely you'd just be building the `cpp-root` container or `base` (in which case you can run `inv docker.build -c cpp-root -c base`. If you think it would be useful then go for it, but I don't think I would ever use it.

CMakeLists.txt
Shillaker: So can we remove the llvm 9 playbook too and have cpp-root use 10 as well?

mfournial: I'm not 100% sure `libWAVM` works with `llvm-10` despite the comment. I thought it worked but when building llvm for perf I ran into issues with it not linking properly, it might just be a bug because my system was giving contradicting information but I wanted to play it safe in the docker build to not have you grinding your teeth when a CI later fails because `libWAVM` didn't link properly. Happy to change though if you think llvm-10 is safe.

Shillaker: Nah, that's fair enough, very wise to play it safe. I had a look through the WAVM commit history and saw no evidence of any LLVM 10-related tweaks, so they've probably not got round to it yet. Should be good to go now.

mfournial: I rebuilt the containers with LLVM-10 and it works, I also tried to rebuild llvm perf with LLVM-10 and it's indeed still that one that doesn't work. As such I will not update to 10 in my changes to the llvm_perf script in #224 but have updated `cpp-root`

Shillaker:
Nice one, thanks for persevering with the LLVM10 stuff, odd about the perf build, but I think we can consider that non-essential for now. I'll merge everything in flight and do a new release today.

PR #230, opened 2020-04-24
Implement multitenancy withh bare metal cluster using simlinks

mfournial:
Mainly add a `faasm` group on the `koala` machines and fix the various issues with this.

The main issue has to do with the permission bit that git track (including in the submodules!) which makes it very hard to sneak some perm changes but I think what's in here should not break the operations downstream and still allows to clone repos and track branches automatically by setting them in `code.yml` or manually (see below). But do watch out if you suddenly have permission issues, especially in the build dir. 

That PR still allow to run git distributed commands too without permission issues like:
```
pdsh -w koala[5,7,8,10].doc.res.ic.ac.uk "cd /usr/local/code/faasm && git checkout -- . && git submodule foreach --recursive git reset --hard"
```

### DIFF
--- a/ansible/roles/code/tasks/main.yml
+++ b/ansible/roles/code/tasks/main.yml
@@ -9,7 +9,7 @@
 
 - name: "Check out code"
   git:
-    repo: "{{ hostvars[host]['remote-url'] | default('https://github.com/lsds/Faasm.git') }}"
+    repo: "{{ hostvars['all']['remote_url'] | default('https://github.com/lsds/Faasm.git') }}"
     dest: "{{ ansible_env.HOME }}/faasm-bm"
     version: "{{ hostvars[host]['version'] | default('master') }}"
     update: yes
--- a/ansible/roles/code/tasks/main.yml
+++ b/ansible/roles/code/tasks/main.yml
@@ -23,6 +23,7 @@
     state: directory
     group: "faasm"
     mode: 0775
+    recurse: yes
 
 - name: "Install system deps"
   become: yes
@@ -38,15 +39,18 @@
     dest: /usr/local/code/faasm
     update: yes
 
-- name: "Directory ownership"
+# Not supported by ansible: recursively change ownership for dirs and files
+- name: "Set directory ownership to faasm"
   become: yes
-  file:
-    path: "/usr/local/code/faasm"
-    group: "faasm"
-    recurse: yes
+  command: |
+    chown -c -R :faasm "{{ item }}"
+  register: chown_status
+  changed_when: chown_status.stdout != ""
+  with_items:
+    - /usr/local/code/faasm
 
 # Not supported by ansible: sets directories to 775 and files to 664
-- name: "Directory permission"
+- name: "Set code permissions to faasm"
   become: yes
   command: |
     chmod -c -R ug=rw,o=r,a-x+X "{{ item }}"
--- a/ansible/roles/code/tasks/main.yml
+++ b/ansible/roles/code/tasks/main.yml
@@ -49,10 +49,11 @@
   args:
     warn: no
 
+# Not supported by ansible: Set small group permission to not trigger git
 - name: "Set code permissions to faasm"
   become: yes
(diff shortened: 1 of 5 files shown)


mfournial:
@Shillaker Oh and I added `koala8` as a worker too. `/usr/local/code/faasm` is simlinked to `/usr/local/code/faasm-git` which is a normal git clone of the repo, but that allows me to swap it out for my own repo when I'm doing remote development

Shillaker:
This is a good plan, but has our college-specific usernames hardcoded in the global role definition, hence it is incorrect for any bare-metal deploy where those usenames don't exist (e.g. if I want to deploy on a machine without the `mfournial` user on it, or I'm working on cloud VMs where I have a different username). In a broader sense, this repo is not specific to LSDS and needs to be deployable by anyone on any machines. To that end, could you avoid referring to LSDS-specific hostnames in the PRs too? 

Any Ansible config that's deployment-specific needs to be added in inventory or variable files in a `.gitignored` directory (e.g. the `ansible/inventory` directory). This allows the files to sit there as they would normally, and people can add exclusions in the .gitignore files if they want to commit them to their forks.

In this instance, I'd probably inline this group as a [host variable](https://docs.ansible.com/ansible/latest/user_guide/intro_inventory.html#assigning-a-variable-to-one-machine-host-variables) in the `bare_metal.yml` inventory file, e.g.

```
[all]
myhost  faasmgroup=user_a,user_b
```

Then in the playbook, the group creation task can use this list _if_ it's present, otherwise it'll just default to the user running the playbook (`ansible_env.USER`). Then just a little note in the bare metal set-up docs and it's good to go.

ansible/roles/code/tasks/main.yml
```diff
@@ -20,11 +37,26 @@
...
-    mode: 0755
\ No newline at end of file
+  command: |
+    chown -c -R :faasm "{{ item }}"
+  register: chown_status
+  changed_when: chown_status.stdout != ""
+  with_items:
+    - /usr/local/code/faasm
+  args:
+    warn: no
```
Shillaker, line 50: Wow, this is grim, but understandable. Is there some gitconfig somewhere to ignore permission changes?

mfournial: I tried this exact thing, unfortunately changing the git config doesn't work for submodules, even when made `--global` or `--system`. Plus it is heavily recommended not to to touch this configuration, especially not to set it user-wide or system-wide.

The best solution in my mind was to create a `faasm` user on which we do everything. This is me trying not to have to go down to such extend because that could break some of the stuff you have so far and can become annoying for containers and stuff, but I think we might have to at some point.

Let's say we try this out and if it doesn't work I'll fix it, as this seems to work fine so far. I'm scared python, `/usr/local/faasm` and everything not on `/usr/local/code/faasm` might be an issue though at some point.

mfournial:
@Shillaker I think I found a way better solution here that I was already doing for a while without really thinking about it:

Each user has a repo sitting at `$HOME/faasm-bm` which when he/she wants to use the cluster gets simlinked to `/usr/local/code/faasm`. That way the user has got the permissions needed to do whatever he wants during the use of the cluster without too much hackery and without the risk of overwriting somebody else's work.

When switching user:
* A user logs in onto his favorite machine and works on his `$HOME/faasm-bm` repo (unmodified because other users do not have access to it). He can even do remote development or launch builds as another user is using the cluster at this point. When he wants to take control and set the cluster to a certain repo state, he will modify the inventory (example below) to include the remote and version he wants) and launch the deployment from there `inv bm.deploy --inventory ~/faasm-bm/ansible/inventory.yml`). This will swap the simlink on all the machines and proceed as usual without group hackery

```ini
[all]
...

[all:vars]
remote_url=https://github.com/mfournial/Faasm.git
repo_version=fix-bm

[worker]
[upload]
[edge]
[redis]
...
```

Shillaker:
Nice, that's very clean and tidy. Good stuff!

PR #232, opened 2020-04-24
Version bump

Shillaker:


mfournial:
Lib-cpp and anything including the main Faasm header will not compile in this release if you don't fix the broken macro at
https://github.com/lsds/Faasm/blob/5ed4b058aaf85f7080b4df04a330359fd7fd6a00/libs/cpp/faasm/faasm.h#L9

PR #233, opened 2020-04-25
Fixes toolchain for the release

mfournial:
Sorry it's my fault, I knew that setting `/usr/bin/clang` sounded wrong on `compiler-rt` and `libcxx` yet it worked the first time I rebuilt the toolchain somehow...

I also used this to address the comment that I left on the PR


Shillaker:
LGTM, thanks for spotting the header snafu too.

PR #237, opened 2020-04-28
Moving more experiment code

Shillaker:


docker/cpp-root.dockerfile
```diff
@@ -3,9 +3,14 @@ FROM ubuntu:18.04
 RUN apt-get update
 RUN apt-get install -y software-properties-common sudo wget
 
-# Note - we have to stick with stable LLVM here for now
+# Note use stable LLVM
 RUN wget -O - https://apt.llvm.org/llvm-snapshot.gpg.key | apt-key add -
 RUN add-apt-repository "deb http://apt.llvm.org/bionic/ llvm-toolchain-bionic-10 main"
+
+# Ansible repo
+RUN apt-add-repository -y ppa:ansible/ansible
```
mfournial, line 11: If you're installing ansible here, why not remove the LLVM-10 code from here that has to be manually maintained against the `llvm-10` playbook?

Shillaker: Yes good point.

PR #238, opened 2020-04-28
Distributed OpenMP experiment 1

mfournial:


include/wavm/openmp/Level.h
```diff
@@ -24,8 +24,6 @@ namespace wasm {
             const int effective_depth = 0; // Number of parallel regions (> 1 thread) above this level
             int max_active_level = 1; // Max number of effective parallel regions allowed from the top
             const int num_threads = 1; // Number of threads of this level
```
mfournial, line 26: I know what you're going to say. The names **are**going to change.

mfournial: It is done on my latest branch

ansible/roles/faasm/templates/worker.conf.j2
```diff
@@ -8,9 +8,9 @@ user={{ ansible_env.USER }}
 environment=LOG_LEVEL=debug,
   FUNCTION_STORAGE=fileserver,
   FILESERVER_URL="http://{{ groups['upload'][0] }}:8002",
-  THREADS_PER_WORKER=20,
```
mfournial, line 11: FYI, having it = to the number of cores in the system creates queuing effects highly visible in experiments. There is always some other process that needs to run while you're trying to run the experiments, so 19 cores will work full on while another one will carry out other tasks, and then there will be one task queuing (if you set 20 tasks for example) that needs for the other cores to free themselves. So I'm being conservative by setting things to 18 and 19

Shillaker: Yeah makes sense, good to know. The bare-metal set-up is really suboptimal here as we need to change this file and can't persist the configuration for more than one experiment at once. Not something to worry about too much now but I'll bear it in mind for the future.

PR #240, opened 2020-05-01
WIP: State reduce

mfournial:
This is an implementation that leaves the native reduce implemention **completely** unmodified, for Faasm compatibility we only require changing the type of the reduce variable from `int64_t` to `i64` which in native is **the same type**:
```c++
using i64 = int64_t
```

The Faasm implementation of this type is as lightweight as possible (although I allowed myself a string in the class, would be better to hold on to a `const char *` since the reduce state key size is fixed). Notably for as minimal impact notably for inside parallel section code using this variable.

The Wasm implementation is actually **entirely in user space** which means that with an emulator for the state it it runs perfectly fine in native too. However why would you use remote distributed state instead of an atomic sum in native (which is why we equate it to `int64_t` instead)


Shillaker:
Nice one, LGTM.

Only other follow up is, as we discussed, adding any useful findings on the OpenMP ABI to the docs, but feel free to do that whenever.

PR #242, opened 2020-05-05
Fix a few lock issues

mfournial:
* Make lockId an optional because of conversion from unsigned int to
long which at worst return a negative number and therefore be treated as
a non set key..! Also it just expressed the intent more cleanly
* Add a local lock to the KV pairs, indeed when a worker has got
multiple threads running it can do boom boom
* Some cleanup

### DIFF
--- a/include/redis/Redis.h
+++ b/include/redis/Redis.h
@@ -123,7 +123,7 @@ namespace redis {
         *  ------ Locking ------
         */
 
-        long acquireLock(const std::string &key, int expirySeconds);
+        std::optional<long> acquireLock(const std::string &key, int expirySeconds);
 
         void releaseLock(const std::string &key, long lockId);
 
--- a/include/redis/Redis.h
+++ b/include/redis/Redis.h
@@ -123,7 +123,7 @@ namespace redis {
         *  ------ Locking ------
         */
 
-        std::optional<long> acquireLock(const std::string &key, int expirySeconds);
+        uint32_t acquireLock(const std::string &key, int expirySeconds);
 
         void releaseLock(const std::string &key, long lockId);
 
--- a/src/redis/Redis.cpp
+++ b/src/redis/Redis.cpp
@@ -423,20 +423,18 @@ namespace redis {
      *  ------ Locking ------
      */
 
-    long Redis::acquireLock(const std::string &key, int expirySeconds) {
+    std::optional<long> Redis::acquireLock(const std::string &key, int expirySeconds) {
         // Implementation of single host redlock algorithm
         // https://redis.io/topics/distlock
-        unsigned int lockId = util::generateGid();
+        long lockId = util::generateGid();
 
         std::string lockKey = key + "_lock";
+        bool result = this->setnxex(lockKey, lockId, expirySeconds);
 
-        long result = this->setnxex(lockKey, lockId, expirySeconds);
-
-        // Return the lock ID if successful, else return -1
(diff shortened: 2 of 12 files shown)


src/state/StateKeyValue.cpp
```diff
@@ -419,21 +419,23 @@ namespace state {
...
 
             if (retryCount >= REMOTE_LOCK_MAX_RETRIES) {
                 logger->error("Timed out waiting for lock on {}", redisKey);
                 break;
             }
 
-            // Sleep for 1ms
-            usleep(1000);
+            // Sleep for 1ms + noise
+            std::chrono::microseconds ms(500 + (rand() % 3000));
```
Shillaker, line 437: Very tiny comment but given microseconds I'd call the variable `us` or `micros`. I've seen too many issues confusing the two before which is why I'm hyper sensitive to it

include/redis/Redis.h
```diff
@@ -123,7 +123,7 @@ namespace redis {
         *  ------ Locking ------
         */
 
-        long acquireLock(const std::string &key, int expirySeconds);
```
Shillaker, line 126: I've not got anything against `std::optional`, but it feels a bit out of sync with the rest of the codebase. Elsewhere I think we take the approach of "return the value if it's worked, and -1 or zero if not". If this is just to avoid the cast from `uint32_t` to `long`, could we just change this return value to `uint32_t` and say `0` means it's failed?

src/redis/Redis.cpp
```diff
@@ -423,20 +423,18 @@ namespace redis {
      *  ------ Locking ------
      */
 
-    long Redis::acquireLock(const std::string &key, int expirySeconds) {
+    std::optional<long> Redis::acquireLock(const std::string &key, int expirySeconds) {
         // Implementation of single host redlock algorithm
         // https://redis.io/topics/distlock
-        unsigned int lockId = util::generateGid();
+        long lockId = util::generateGid();
```
Shillaker, line 429: See point re. changing return value of `acquireLock` to `uint32_t` which this returns.

src/state/RedisStateKeyValue.cpp
```diff
@@ -18,12 +18,17 @@ namespace state {
         return redis.strlen(actualKey);
     }
 
+    // TODO - the remote locking here is quite primitive since we ignore the fact threads can run
+    // on the same machine. Redis is also aware of scheduling and so we could optimise this.
```
Shillaker, line 22: Agreed. Would require some careful thought on how to do without introducing deadlock potential. Let's see how long `RedisStateKeyValue` survives, it may be killed soon anyway.

src/state/StateKeyValue.cpp
mfournial: I'd even prefer using the name constructors like `1ms` but it doesn't work 😕.
I oscillated in my tests between using `milliseconds`, `microseconds` and `nanoseconds` so it's actually a very good catch!

mfournial: Actually, make sure you turn on unicode support, i'm going to call it `μs` 🙃

mfournial: I removed it entirely. randomisation of wait time isn't a great way to do things and may hide issues we'd wanna spot

Shillaker:
LGTM apart from potential compile error in test. Once fixed happy to merge if it's just a case of the flaky MPI test.

PR #245, opened 2020-05-08
Various OpenMP reduce experiment

mfournial:


mfournial:
@Shillaker That should be ready for review. I think you're going to grimace at how I made the value lock for the KV a recursive lock instead of a shared one in order to basically make `lockGlobal` really work. There is a plethora of designs obviously here that we can do to avoid this, I think however this is the simplest one to revert if you need to change it later.

Shillaker:
This is insteresting and I think I understand the motivation, but need to clarify, I have a feeling it comes down to defining the responsibilities of the "global" lock. 

In Faasm's two-tier state system there's a master/ replica model where there's one master copy and many replicas. The point of the "global" lock is _only_ to lock the master copy, _not_ to do any locking on the local replica. If applications need fine-grained control over local locking, they can use direct pointer access and call `lockWrite`/ `lockRead` etc. All the high-level state operations such as get/ set will also perform their own local locking to ensure atomicity. 

The global lock is primarily used to ensure consistency across hosts when modifying the master copy, e.g. you might have: lock global, pull, get, set, push, unlock global. The global lock then creates a sort of distributed cricital section that ensures only one thread in the system will be modifying the state at a given time. This works provided all threads adhere to this approach, i.e. acquiring a global lock before editing the local replica. However, this does still permit a thread to modify the local replica _without_ holding the global lock, if the application is written to allow this. Users can add an extra layer of locking to guard against this in two ways: (i) create another shared state value that is used to do the global-lock-free parallel processing, then atomically copy to/ from this as part of the global cricial section, (ii) add extra nested locking operations, i.e. lock global, _lock write_, pull, get, set, push, _unlock write_, unlock global.

I think pursuing option (ii) has led you to merge `lockWrite`/ `unlockWrite` into `lockGlobal`/ `unlockGlobal` and add the `recursive_mutex`. This is justified as (ii) actually won't work with the current code. Because all state operations perform implicit locking, calling any of them inside `lockWrite`/ `unlockWrite` will break, as they attempt to re-lock on the same mutex. The only thing that doesn't break is using direct pointer access to the state (as this is explicitly lock free). 

To make the existing state operations work inside the local locking, we could make the mutex into a `recursive_mutex` as you have done, but this removes the ability to do a read/ write lock. Having a read/ write lock is important for performance in read-heavy applications, so we can't lose that. I'm also really reluctant to introduce a new class of recursive locking bug just to solve this problem. 

Another solution is to introduce a "lock free" flag to each of these operations, which might end up being the easiest, but pushes more responsibility onto the user. Another solution is to add some custom reentrant behaviour. This will be something like "Do I already have an open local read/ write lock, if so I'll ignore my normal implicit locking". I need to see how grim this would be and whether it can be done with just a couple of atomic bools. I'll get back to you on that one...

mfournial:
Agreed with everything you said, especially re the responsibility of the global lock coming from the Faasm state perspective, I came hard from the other side of making OpenMP serverless without changing too much user code, and showing what important primitives that needed (the pointer access not worker it for this). That's why I used `incrby` initially because a variable with a local mutex is all we need really.  

I wasn't expecting the lock changes to get through anyway, especially with the new work you're doing so I removed it. We could talk about distributed locks and liveness issues all day.

I'm happy with the PR as it it now that this is removed.

mfournial:
Actually I'm going to work from this branch to do the continuity experiment, marking this back as WIP

mfournial:
I pulled the master changes and checked that everything was fine, so fingers crossed all of that works

PR #246, opened 2020-05-08
Implement single pragma

mfournial:
This is the last OpenMP pragma I think we should be supporting that is useful for WasmMP with the full list being:
|  Supported operations |
|-----------------------|
|         atomic        |
|        barrier        |
|        critical       |
|         flush         |
|       for [simd]      |
|         master        |
| parallel (\| section) |
|         single        |

The rest of the pragmas that I know of are either used for teams and tasks stuff (like a thread pool) but that doesn't seem to be used a lot, or there is a ton of new pragmas arriving for `target` devices (e.g. GPUs), but that's work for someone else 😰


Shillaker:
Nice one, LGTM

PR #247, opened 2020-05-08
Implement preallocated thread stack to fix LULESH compatibilty for WasmMP 

mfournial:
Reason of existence of this PR: make LULESH work with `WasmMP`. So there is going to be a few changes for that to happen
* [x] bring back dlmalloc. (not super essential now that there is WASI support as far as I understand the `__wasm_call_ctors`). Since the third party repo was still there (and some demo func existed), I figured the fact that the `inv` task had disappeared from the `libs` and that it wasn't in the release anymore because of maybe a hasty deletion.
* [x] Make the profiling things CMake  options. I wanted to do that for a long time, I happened to profile again so it was the occasion.
* [x] Fix the reason why LULESH was generating a trap at runtime.
* [x] Fix performance issues with LULESH since we are currently 1,000,000 times slowly approximately. There is an `iterations` parameter for lulesh, normally set at 10,000. I have to set it to 1 or 2 max to be able to finish.

Those two final points were fixed implementing a thread pool which at the start of a module allocates thread memory and assign to each thread in the pool this stack to be reused (no need for collection and/or more allocations). This is a good-enough "WasmMP" implementation implementation of the thread pool which us gives a firmer understanding of what should be done for making this compatible with faaslets. Something like having another linear memory, which to my understanding is implemented in WAVM, just for the thread stack could work here for example.. That way the stacks wouldn't have to be cloned, and they can be created in any number, on demand, on each host.

To see just the thread pool changes, I think mfournial#6 gives a clearer view. Below is more explaination on why I had to do those changes for LULESH, and possibly other OpenMP programs.  

### LULESH bug and optimisation

With its >20 pragmas called in loops, LULESH has about a thousand of calls to `fork` when `iterations` = 2, each creating 20 threads (not even talking about nested sections). The way things were implemented meant that each new thread required `mmap`ing a stack that
1. was never collected which means memory pressure was slowly growing before blowing up (**this was the bug**)
1. is really slow, remember we're talking about creating several thousands of threads per iteration (**this is the performance issue**).

Since I don't think we have `munmap`, I tried to tackle the two problems at once by implementing "**Prealloacted Thread stack**". The first level (not its nested levels) pre-allocates on start-up `numThreads x stackSize` giving it `numThreads` pointers to stack tops that it can give to each thread on creation. Those regions (stacks) can be reused as the same level is forked over and over again.

If we don't have nested regions, then the **bug** is fixed, and hopefully the performance bug is gone too. Although the former is true, the latter is not and we're still thousands of times slower than native code. OpenMP is notorious for not being performant when abusing its _fork and join_ mechanism and I thought all code was written with that in mind, but it seems like `libomp` and `libgomp` have made such optimisations to eliminate this problem that code can and has be written that way anyway. Those optimsations are:
1. The single thread optimisation. If a parallel section runs with only one thread. `libomp` devs clearly went out of their way to implement it it seems, so it must be important. But it's not relevant for LULESH that much I think since the effective depth is always 0 (weird it didn't use to be!)
2. Having a thread pool design instead of dynamic thread creation.

_Note_ that since nested section was not activated by LULESH, OpenMP prescribes nested levels have a 1 thread limit. Since we unlike `libomp` still create a thread, this has very high overhead.

### Preallocated thread stack implementation

Preprocessor macros, that's all I could think of doing to have it as a feature flag. It really has to be a feature because it's not a backwards compatible change (regarding nested level experiments), and because of its high start up cost is not something you'd want by default.
That might have to change now that I have to do a thread pool, but I think this will remain.


mfournial:
Comparison of native with WasmMP before the thread pool implementation:

im

mfournial:
A rudimentary thread pool implementation brings us within 2 orders of magnitudes (only/) from native code. To not clutter this PR yet since the code is a draft I have it here: https://github.com/mfournial/Faasm/pull/6

mfournial:
I implemented what I thought to be a better thread pool implementation for the problem, but it ended up 2x slower: https://github.com/mfournial/Faasm/pull/7/files. Documenting in case I want to roll back to it, but I think I'll use the first design

include/wavm/WAVMWasmModule.h
```diff
@@ -112,6 +114,10 @@ namespace wasm {
 
         int getDataOffsetFromGOT(const std::string &name);
 
+        U32 allocateThreadStack();
+
+        std::unique_ptr<PlatformThreadPool> &getPool();
```
Shillaker, line 119: As the current thread pool stuff is all OMP-specific, can you make that clear in all the names? It will avoid confusion with the pthread support. This would be `getOMPThreadPool()` or something along those lines.

src/wavm/CMakeLists.txt
```diff
@@ -24,6 +25,7 @@ set(LIB_FILES
         mpi.cpp
         network.cpp
         openmp.cpp
+        PlatformThreadPool.cpp
```
Shillaker, line 28: Can you rename `PlatformThreadPool` to `OMPThreadPool` as per the comment above re. it being OMP-specific?

src/wavm/PlatformThreadPool.cpp
```diff
@@ -0,0 +1,80 @@
+#include "PlatformThreadPool.h"
+
+#include <wavm/openmp/ThreadState.h>
+#include <wavm/openmp/openmp.h>
```
Shillaker, line 4: Would be nice to make this non-OMP specific if possible, but probably unavoidable atm/ not worth it, so just rename as per other comments.

mfournial: Making the ThreadPool non OMP specific is too much of a game of chicken and eggs, the maze of headers is already very complex, I try to have everything in `wavm/openmp` not known about `wavm` to avoid cycles. But the `PlatformThreadPool` being OpenMP specific, I think it clearly belongs there indeed. I don't see a pretty way to do this though, I could move the `PlatformThreadPool` files inside the `openmp` folder, but build it along the `wavm` folder, and not the `wavm/openmp` lib. I'd find this confusing.

mfournial: I placed it in the openmp namespace actually, a lot more explicit probably

src/util/config.cpp
```diff
@@ -53,6 +53,7 @@ namespace util {
         maxInFlightRatio = this->getSystemConfIntParam("MAX_IN_FLIGHT_RATIO", "3");
         maxWorkersPerFunction = this->getSystemConfIntParam("MAX_WORKERS_PER_FUNCTION", "10");
         threadMode = getEnvVar("THREAD_MODE", "local");
+        ompThreadPoolSize = this->getSystemConfIntParam("OMP_THREAD_POOL_SIZE", "10");
```
mfournial, line 56: Set to 0 for no overhead on module start (I set it to 10 here because it needs to be on for tests, I could set it to 0 by default and change the env var in the CI instead

Shillaker: Yes, I think the default configuration should be "all bells and whistles switched off", then they can be turned on when necessary.

Shillaker:
Nice one. Thanks for being so patient and making all those changes, much appreciated.

PR #264, opened 2020-07-10
SGX tidy-up part 2

Shillaker:


include/sgx/SGXWAMRWasmModule.h
```diff
@@ -8,37 +8,63 @@
...
-        bool execute(message::Message &msg, bool force_noop = false);
+        bool execute(message::Message &msg, bool forceNoop = false);
 
         const bool isBound();
+
     private:
-        bool _is_bound = false;
-        unsigned int thread_id;
-        std::vector<uint8_t> wasm_opcode;
+        bool _isBound = false;
```
Shillaker, line 64: @J-Heinemann for better or worse we use camel case in the rest of the project, so top-level headers and anything that might interface with the rest of the system should follow this convention. For module-private implementations and lower level code we don't need to worry too much, although obviously it's good to maintain consistency where possible.

PR #294, opened 2020-09-13
WAMR Interpreter mode added and minor bug fixes

J-Heinemann:


CMakeLists.txt
```diff
@@ -10,8 +10,8 @@ option(FAASM_OPENMP_TRACING "Trace OpenMP forks in Redis" OFF)
 option(FAASM_PERF_PROFILING "Turn on profiling features as described in debugging.md" OFF)
 
 # SGX functionality
-option(FAASM_SGX_SUPPORT "Add support for SGX" OFF)
-option(FAASM_SGX_SIM_MODE "SGX sim mode flag" OFF)
+set(FAASM_SGX_SUPPORT 0) # Enable/Disable SGX Support
```
Shillaker, line 13: All on/ off switches should be expressed as `option`s. This way people can pass them as arguments on the command line when configuring, rather than having to change this file. 

Importantly they can then run `cmake -LH` which lists all the project options and their associated help text. If things are set as variables they don't show in this list.

CMakeLists.txt
```diff
@@ -75,6 +76,7 @@ set(WAMR_ROOT_DIR ${CMAKE_CURRENT_LIST_DIR}/third-party/wamr)
 set(WAMR_SHARED_DIR ${WAMR_ROOT_DIR}/core/shared)
 set(WAMR_IWASM_DIR ${WAMR_ROOT_DIR}/core/iwasm)
 set(WAMR_BUILD_TARGET X86_64)
+set(WAMR_EXECUTION_MODE AoT) # Select the WAMR execution Mode (AoT/Interpreter)
```
Shillaker, line 79: This is great, but should also be set as a project option at the top for the same reasons as `FAASM_SGX_SUPPORT`. As there are only two options, I'd set it as:

```
option(WAMR_INTERPRETER_MODE "Turns interpreter on/ off (vs. AoT)" OFF)
```

This will simplify the `if` statement below.

CMakeLists.txt
```diff
@@ -89,6 +91,13 @@ include_directories(
         ${WAMR_IWASM_DIR}/common
 )
 
+#WAMR Interpreter / AoT Switch
+if(WAMR_EXECUTION_MODE STREQUAL "Interpreter")
```
Shillaker, line 95: With the interpreter mode as an option, it can change to:

```
if(WAMR_INTERPRETER_MODE)
...
```

This avoids the need to do string matching.

include/wamr/WAMRWasmModule.h
```diff
@@ -42,7 +43,11 @@ namespace wasm {
     private:
         bool _isBound;
 
-        std::vector<char> errorBuffer;
+        char errorBuffer[ERROR_BUFFER_SIZE];
+
+#if(WAMR_EXECUTION_MODE_INTERP)
+        std::vector<uint8_t> wasmBytes;
```
Shillaker, line 49: Why does this need to be a property of the class? AFAICT it's only used in one method (`bindToFunction`), so it can just be a local variable in that method.

src/wamr/CMakeLists.txt
```diff
@@ -82,7 +88,12 @@ llvm_map_components_to_libnames(
 faasm_private_lib(wamrmodule "${LIB_FILES}")
 target_link_libraries(wamrmodule
         wasm
-        aotclib
         wamrlib
-        ${WAMR_LLVM_LIBRARIES}
         )
+if(WAMR_EXECUTION_MODE STREQUAL "AoT")
```
Shillaker, line 93: See comment above re. making this an option.

src/wamr/WAMRWasmModule.cpp
```diff
@@ -69,35 +75,38 @@ namespace wasm {
 
         // Load the wasm file
         storage::FileLoader &functionLoader = storage::getFileLoader();
-        std::vector<uint8_t> aotFileBytes = functionLoader.loadFunctionWamrAotFile(msg);
+#if(WAMR_EXECUTION_MODE_INTERP)
+        wasmBytes = functionLoader.loadFunctionWasm(msg);
+#else
+        std::vector<uint8_t> wasmBytes = functionLoader.loadFunctionWamrAotFile(msg);
+#endif
```
Shillaker, line 82: See comment above, why can't this be a local variable in `WAMR_EXECUTION_MODE_INTERP` too?

src/wamr/WAMRWasmModule.cpp
```diff
@@ -69,35 +75,38 @@ namespace wasm {
...
-            logger->error("Failed to instantiate WAMR module: \n{}", errorMsg);
-            throw std::runtime_error("Failed to instantiate WAMR module");
+            std::string errorMsg = std::string(errorBuffer);
+            logger->error("Failed to load WAMR module: \n{}", errorMsg);
+            throw std::runtime_error("Failed to load WAMR module");
         }
         
         // Instantiate module
-        moduleInstance = wasm_runtime_instantiate(
+        if(!(moduleInstance = wasm_runtime_instantiate(
```
Shillaker, line 99: In this project we don't nest function calls in if statement conditions, rather split them onto separate lines:

```
moduleInstance = wasm_runtime_instantiate(
   ...
);

if(!moduleInstance) {
...
}
```

Shillaker (changes requested):
Great, thanks for this Joshua!

A few small style/ config comments but otherwise looks good.

I'm not sure why there are changes to `third-party/faabric` and `third-party/pyodide` though. Could you get those submodules in sync with master?

CMakeLists.txt
J-Heinemann: Sounds good for me.

include/wamr/WAMRWasmModule.h
J-Heinemann: Unfortunately not, because wasm_runtime_load creates the WASMModule but doesn't copy the code into this module, hence you have to keep the code somewhere else during the whole execution. 
This solved the bug (I mentioned on slack) that the first execution works fine in interpreter but the next ones will fail.

src/wamr/CMakeLists.txt
J-Heinemann: Sounds good 👍

src/wamr/WAMRWasmModule.cpp
J-Heinemann: 👍

PR #308, opened 2020-09-29
Add CLAPACK support

Shillaker:
Closes #308


faasmcli/faasmcli/tasks/libs.py
```diff
@@ -66,15 +62,9 @@ def native(ctx, clean=False):
...
-        raise RuntimeError("Failed to build native tools")
-
-    res = call("ninja", shell=True, cwd=build_dir)
-    if res != 0:
-        raise RuntimeError("Failed to make native tools")
-
-    call("sudo ninja install", shell=True, cwd=build_dir)
+    run(build_cmd_str, shell=True, cwd=build_dir, check=True)
+    run("ninja", shell=True, cwd=build_dir, check=True)
+    run("sudo ninja install", shell=True, cwd=build_dir, check=True)
```
Shillaker, line 67: Changed all calls to `call` and `check_output` to the more modern (and concise) `run`. Considered a separate PR but not sure it's worth it

faasmcli/faasmcli/tasks/libs.py
```diff
@@ -231,57 +218,43 @@ def eigen(ctx, verbose=False):
...
+        run("make clean", cwd=lapacke_dir, shell=True, check=True)
+
+    n_cpu = int(cpu_count()) - 1
+
+    # Make libf2c first (needed by others)
+    run(
+        "make f2clib -j {}".format(n_cpu),
+        shell=True,
+        cwd=clapack_dir,
+        check=True,
```
Shillaker, line 250: This is new. `zlib` and `libpng` were removed a while ago but I missed these tasks

src/wavm/io.cpp
```diff
@@ -665,6 +666,15 @@ namespace wasm {
...
 
+    WAVM_DEFINE_INTRINSIC_FUNCTION(wasi, "fd_renumber", I32, wasi_fd_renumber, 
+            I32 fdOld, I32 fdNew) {
+        throwException(Runtime::ExceptionTypes::calledUnimplementedIntrinsic);
+    }
+
+    WAVM_DEFINE_INTRINSIC_FUNCTION(env, "tmpfile", I32, tmpfile) {
+        throwException(Runtime::ExceptionTypes::calledUnimplementedIntrinsic);
+    }
+
```
Shillaker, line 677: @csegarragonz we often have to add functions like this to keep the linker happy but they're never actually called at runtime (and they'll throw an exception if they are, at which point we can think about implementing them).

PR #313, opened 2020-10-01
Refactor URLs from lsds/faasm to faasm/faasm

csegarragonz:
The title is pretty self-explainatory. Just changed the pointers after the ownership's transfer.


csegarragonz:
There are still some pointers in third-party modules, but only in the READMEs, I won't create a PR for those then.

Shillaker (approved):
Nice :+1:

Issue #302, opened 2020-09-28
Add Horovod + MXNet support


Shillaker:
We'll target Horovod and MXNet as per [this example](https://github.com/horovod/horovod/blob/master/examples/mxnet/mxnet_mnist.py). 

- [ ] Ensure numpy using correct CBLAS and LAPACK
- [ ] Build MXNet and downstream deps to wasm shared libs
- [ ] Run Horovod through standard cross-env build (https://github.com/faasm/faasm-cpython)
- [ ] Add Faasm Python function to run code similar to Horovod MXNet example

PR #315, opened 2020-10-05
Improvements to shared/ static module builds and debugging

Shillaker:
This moves a lot of the compiler flags and configuration around building shared libraries upstream into this repo. 

This avoids such information being spread over multiple files, but it's still duplicated in this project and the `faasm-toolchain`. Another piece of work will be to unify the two and move all this config into `faasm-toolchain`.

Part of #302


faasmcli/faasmcli/tasks/libs.py
```diff
@@ -227,33 +227,3 @@ def eigen(ctx, verbose=False):
         check=True,
     )
 
-
```
Shillaker, line 230: The BLAS stuff has been moved to the `faasm-clapack` submodule

func/dynlink/CMakeLists.txt
```diff
@@ -5,36 +5,5 @@ set(BUILD_SHARED_LIBS ON)
 
 include_directories(include)
 
-# This is critical otherwise the linker tries to statically link in standard
```
Shillaker, line 8: This configuration is all now handled by the toolchain

src/wavm/codegen.cpp
```diff
@@ -6,13 +6,17 @@
...
 #include <WAVM/WASTParse/WASTParse.h>
 
+#include <faabric/util/logging.h>
 #include <faabric/util/files.h>
 
 using namespace WAVM;
 
 namespace wasm {
-std::vector<uint8_t> wavmCodegen(std::vector<uint8_t>& bytes)
+std::vector<uint8_t> wavmCodegen(std::vector<uint8_t>& bytes,
```
csegarragonz, line 15: So these changes to `doCodegen` and `wavmCodegen` are for better error reporting I guess? It's alright, just a bit confusing that they are included in this PR.

csegarragonz (approved):

csegarragonz:
LGTM

src/wavm/codegen.cpp
Shillaker: I guess it depends on how strict we want to be with single purpose PRs. Given how long the build takes at the moment I think it's OK to include small improvements like this alongside other changes.

Shillaker: I take your point that the title is kind of misleading. Although it's updating the submodules that contain changes related to Horovod/ MXNet, the changes in this repo are more generally relevant to building libraries. I've renamed the PR to better reflect the fact that it's a bit of a catch-all

PR #317, opened 2020-10-06
Faasm support to run LAMMPS

csegarragonz:
Related to faasm/faasm-experiments#13

**Steps towards running Lammps with Faasm:**
* [x] Run `lammps` locally.
* [x] Run `lammps` natively on the cluster using MPI.
* [x] Corss-compile `lammps` with `faasm-toolchain`.
* [x] Run `lammps` over `faasm`.

PS: I have not formatted the code as I was going through it. However, there's an issue with how JetBrains related IDEs handle newlines that leaves a trace of breadcrumbs through all the files that I open and modify. I only not-restore the ones that I am actually changing. At some point we'll have to refactor the whole codebase with the new `clang-format` :man_shrugging:


Shillaker:
> at some point we'll have to refactor the whole codebase with the new `clang-format` man_shrugging

Unfortunately we still have to wait for a couple of downstream bits of work until we can reformat the whole thing with `clang-format` :disappointed: , let's discuss where you're seeing the newlines though, because I think I've already formatted most of the parts you should be editing for this piece of work...

csegarragonz:
This branch is now stale, as most of the contributions have already been pushed upstream, and it will render useless as we merge to the new experiment structure repositories.

PR #320, opened 2020-10-07
Manually adding newlines to the end of a file without one, as clang does not enforce it from the shell.

### LINKED ISSUE #318: Clang-format newlines at EOF
Enforce a consistent style on this (either with or without, but it should be impossible to pass the check if the file has the wrong ending).

Note that the check that we perform in CI will be with [`run_clang_format.sh`](https://github.com/faasm/faasm/blob/master/bin/run_clang_format.sh).

csegarragonz:
@Shillaker testing the script seems it works fine (a clear case of N=1, where N is the number of samples). Should we also run it before merging the PR? I'll let you decide.

Closes #318


Shillaker (approved):
Let's just merge this in for now, then all the remaining formatting changes will be captured in the full-repo-formatting PR

PR #322, opened 2020-10-13
Update to self-contained toolchain

Shillaker:
This change has two main parts:

_Containerisation_

- Use containerised builds for `faasm-toolchain` and `faasm-cpython` (removing third-party, and removing need to package artifacts with Faasm releases)
- Dockerise the Faasm CLI so users can just pull the image and start developing
- Rename existing `toolchain` container to `cli`
- Use the `cli` container to run the tests (means we can remove `base-test` and get the benefit of regularly checking `cli` is working)
- Build and push containers with Github Actions when a tag is created

_Third-party deps_

- Remove as many as possible (using CMake `ExternalProject` or `FetchContent`, or by moving to other repos)
- Move all toolchain utilities in this repo into the `faasmtools` library in the toolchain repo

Unfortunately we have to keep `wamr` and `xra` as Git submodules for now as there is still some work outstanding that makes heavy use of these.


.github/workflows/release.yml
```diff
@@ -0,0 +1,68 @@
+name: Release
```
Shillaker, line 1: This builds all the containers needed for a Faasm release. It'll be slow, but it's a start

docker-compose-cli.yml
```diff
@@ -0,0 +1,25 @@
+version: "3"
+
```
Shillaker, line 2: This is used to initialise the CLI with a containerised redis instance, but could include more stuff in the future (e.g. mocked object storage)

docker-compose.yml
```diff
@@ -15,7 +15,8 @@ services:
       - redis-queue
       - redis-state
     volumes:
-      - ./machine-code/:/usr/local/faasm/object/
+      - ./container/machine-code/:/usr/local/faasm/object/
+      - ./container/shared_store:/usr/local/faasm/shared_store
```
Shillaker, line 19: This means we don't have to keep "uploading" the Python functions

faasmcli/faasmcli/tasks/docker_tasks.py
```diff
@@ -4,45 +4,30 @@
...
-from subprocess import call, check_output
+from subprocess import run, PIPE
 
-import jinja2
 from invoke import task
 
 from faasmcli.util.env import PROJ_ROOT
 from faasmcli.util.version import get_faasm_version
 
-# Order matters here
```
Shillaker, line 15: These Docker tasks are a lot simpler now because a) the actual build/ release part is done in Github; b) we no longer have container-specific dockerignore files

faasmcli/faasmcli/tasks/libs.py
```diff
@@ -113,14 +108,6 @@ def faasmpi(ctx, clean=False, verbose=False):
...
     Compile and install the Faasm MPI library
     """
-    # Build the Faabric MPI target
-    _build_faasm_lib(
-        "third-party/faabric/src/mpi",
-        clean,
-        verbose,
-    )
-
-    # Build the Faasm wasm wrapper
```
Shillaker, line 123: This build of the Faabric MPI stuff is now captured in the CMake files

faasmcli/faasmcli/tasks/libs.py
```diff
@@ -191,38 +176,3 @@ def fake(ctx, clean=False):
     for so in shared_objs:
         print("Running codegen for {}".format(so))
         run("{} {}".format(binary, so), shell=True, check=True)
-
-
```
Shillaker, line 195: Eigen is now build as part of the normal sysroot

faasmcli/faasmcli/tasks/python.py
```diff
@@ -34,72 +26,6 @@ def _remove_runtime_dir(dir_name):
         rmtree(dir_path)
 
 
-@task
```
Shillaker, line 37: CPython stuff is now handled in the faasm-cpython repo

faasmcli/faasmcli/tasks/release.py
```diff
@@ -163,43 +167,6 @@ def create(ctx):
     )
 
 
-@task
-def upload(ctx, which=None):
```
Shillaker, line 167: We no longer bundle artifacts with the Faasm release, but instead with the relevant container build for the other repos

faasmcli/faasmcli/tasks/wast.py
```diff
@@ -3,36 +3,7 @@
...
-from faasmcli.util.env import PROJ_ROOT, THIRD_PARTY_DIR
-from faasmcli.util.files import clean_dir
-
-WABT_DIR = join(THIRD_PARTY_DIR, "wabt")
-WABT_BUILD_DIR = join(WABT_DIR, "build")
-WASM2WAT_BIN = join(WABT_BUILD_DIR, "wasm2wat")
-DECOMPILE_BIN = join(WABT_BUILD_DIR, "wasm-decompile")
-
-
-@task
```
Shillaker, line 16: WABT is now installed as part of the CLI container

faasmcli/faasmcli/tasks/toolchain.py
```diff
@@ -1,187 +0,0 @@
-from os import makedirs
```
Shillaker, line 1: Toolchain is now built in the faasm-toolchain repo

faasmcli/faasmcli/util/env.py
```diff
@@ -47,30 +55,8 @@
...
 
-FAASM_TOOLCHAIN_FILE = join(
-    PROJ_ROOT, "third-party", "faasm-toolchain", "WasiToolchain.cmake"
-)
-FAASM_SYSROOT = join(FAASM_LOCAL_DIR, "llvm-sysroot")
-SYSROOT_INSTALL_PREFIX = FAASM_SYSROOT
-
 LATEST_CMAKE = "/usr/local/lib/cmake-3.15/bin/cmake"
 
-POSSIBLE_BUILD_BINS = [
```
Shillaker, line 58: This was all about detecting different environments which is unnecessary now

faasmcli/faasmcli/util/release.py
```diff
@@ -1,103 +0,0 @@
-from os.path import join
-from subprocess import check_output
-
-from faasmcli.util.env import FAASM_LOCAL_DIR, FAASM_TOOLCHAIN_FILE
-from faasmcli.util.version import get_faasm_version
-
-TOOLCHAIN_INSTALL = join(FAASM_LOCAL_DIR, "toolchain")
-
-
-def _get_artifact_url(name, version=None):
```
Shillaker, line 10: Same story re. artifacts no longer being bundled

faasmcli/faasmcli/util/toolchain.py
```diff
@@ -1,149 +0,0 @@
-from copy import copy
-from os.path import join
```
Shillaker, line 2: All this info is in the faasm-toolchain repo now

docs/development.md
```diff
@@ -2,136 +2,103 @@
...
+### Tooling - editors, IDEs etc.
 
-Faasm depends on Protobuf and gRPC which are set up with the Ansible `grpc` role
-found [here](ansible/roles/grpc/tasks/main.yml). However, if you want to intsall
-these yourself or use an existing version, make sure you use the full 
-[gRPC C++ build](https://grpc.io/docs/languages/cpp/quickstart/).
+Several directories (including the project root) are mounted as volumes in the
+CLI container (see the [CLI `docker-compose.yml`](../docker-compose-cli.yml)). 
+This means you can edit the files directly on your host and compile them inside 
+the container.
```
csegarragonz, line 51: There's a problem with this approach, and it's the old [root/uid problem](https://github.com/moby/moby/issues/3124#issuecomment-104937065), you can also see this in their [docs](https://docs.docker.com/compose/compose-file/#configs). In particular, from outside the container, all the artifacts generated from _within_ the container, will be owned by root. I would personally use the same `uid:gid` as the user invoking the script, hence from outside the container we would not see a difference in the generated builds.

We could either modify the underlying `cli.dockerfile` or `docker-compose-cli.yml`, I think the latter is cleaner as we can easily query  parameters from the command line. I'd rather centralize my suggestions in this comment than in the respective files.

_After working for an hour on it_
I have some ideas on how to do this, and would be willing to do a PR myself. It's not trivial though. We could discuss this personally.

docs/development.md
```diff
@@ -2,136 +2,103 @@
...
+To build and run the tests, you can then run the following inside the container:
 
 ```bash
-# Python stuff and other dependencies
-sudo apt install python3-dev python3-pip python3-venv libcairo2-dev libtinfo5
-sudo pip3 install -U pip
+# Build the tests and the codegen targets
+inv dev.cmake
+inv dev.cc tests
+inv dev.cc codegen_func
```
csegarragonz, line 30: Isn't this done in the `cli.dockerfile` lines 37 and 38?

csegarragonz:
This is great stuff! I have checked out the code in a separate location and fiddled around a bit see if I could break stuff.

I have tried running the installation process in two different machines, and in both of them it seems to fail. In particular I have run it in a clean directory (clean clone) on my machine, and also in `koala5`. In both cases everything goes fine until I launch the `tests` command. In which case I receive the following:
```bash
terminate called after throwing an instance of 'std::runtime_error'
    what(): could not resolve host redis
Aborted (core dumped)
```

csegarragonz (changes requested):

docker-compose-cli.yml
```diff
@@ -0,0 +1,25 @@
+version: "3"
```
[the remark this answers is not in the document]
Shillaker: If we put the normal `docker-compose.yml` file in the `docker` directory, `docker-compose` always requires `-f docker/docker-compose.yml`, which is a bit annoying. I can put the `docker-compose-cli.yml` one in there though.

docs/development.md
Shillaker: Yes I'm aware of this, and although it can occasionally be inconvenient, how much of a problem is it in this case?

The directories used for volumes are either implicitly read only (the sources), or created specifically for mounting (i.e. `build` or `container/*`). Neither `build` nor `container` are checked in, and can be owned completely by the container, so this shouldn't cause any conflicts. The only issue I can see here is if you create new source files from within the container, which could become a problem.

docs/development.md
Shillaker: Yes, but when using the CLI for development we mount our own `build` directory which overwrites the stuff that's built here. This is so that build artifacts are kept on the local machine and don't have to be regenerated when you recreate your CLI container. 

These _are_ used when we use the CLI for running tests though.

docs/development.md
```diff
@@ -2,136 +2,103 @@
...
-For now, many locations rely on the code being located at
-`/usr/local/code/faasm` (note the lowercase "f"), with the latter two 
-directories _owned_ by the current user. This is annoying and something we will 
-fix in future. 
-
-You can either set this directory up directly, or just symlink it.
-
-Assuming you've checked out this code somewhere, you'll need to make sure
-submodules are up to date:
+Most development can be done using the containerised CLI. To get everything set
```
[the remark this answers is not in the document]
Shillaker: I personally would have thought this was obvious, and is probably overkill given that these docs are aimed at people wanting to do development.

A more obvious place to mention this would be in the main `README.md`, where the quick-start assumes people can run docker-compose. Again, I would have thought this was obvious, but if you want to include the links then perhaps put them on there.

Shillaker:
> This is great stuff! I have checked out the code in a separate location and fiddled around a bit see if I could break stuff.
> 
> I have tried running the installation process in two different machines, and in both of them it seems to fail. In particular I have run it in a clean directory (clean clone) on my machine, and also in `koala5`. In both cases everything goes fine until I launch the `tests` command. In which case I receive the following:
> 
> ```shell
> terminate called after throwing an instance of 'std::runtime_error'
>     what(): could not resolve host redis
> Aborted (core dumped)
> ```

I've not seen this before, is the Redis container running? To check which containers are running you can run `docker ps`

docs/development.md
csegarragonz: As we discussed, I agree on your point, and I see that _practically_ this is not an issue. However, I consider that installing a piece of software and, as an artifact, ending up with `root`-owned (noun phrase) directories in your home folder to be far from desirable.

I also understand that this is somewhat unrelated to this PR, and the container-based-dev will keep evolving, so I move the discussion to #328.

csegarragonz:
Closes #325

csegarragonz:
The `redis` problem is indeed solved but, just to double-check, do all tests pass in your build? They seem to pass in the GH action, but they fail on my machine. In particular, the RPC test fails reporting:
```bash
RPC error state_size: failed to connect to all addresses
```
In particular this is located at
```bash
/usr/local/code/faasm/tests/utils/worker_utils.cpp:152
```

Shillaker:
Hmm, they seem to work on my machine too. Given that it works in GH I think it's best to merge this and raise another issue if we see it happening regularly.

Shillaker:
Closes https://github.com/faasm/faasm/issues/314

csegarragonz:
Okey I'll add an issue for it and assign it to myself #329.

Other than that, LGTM and happy to merge the big boy.

csegarragonz (approved):

PR #326, opened 2020-10-21
Pass over MPI Tests

csegarragonz:
In this PR I go through the MPI tests and include some style homogenization. I have added some checks that were missing, removed some redundant others, and tried to enforce a uniform style on pointers (I have used the one that appeared more frequently, i.e. `new` + `delete`, and changed where the `C`-like one was used.

As I went through the files, I also linted them as some were missing the newline. I know I said it was better not to mix these two things. Feel free to suggest including them in a separate PR (or leave them for the future).

I have also included an ansible role for the `openmpi` playbook (which was missing). And include two necessary workarounds to get `catch2`'s playbook working. These will be overwritten by #322, but in the meantime it's preferable to have the system in a working state (hence why you should approve the hacky workaround, he he).


ansible/catch.yml
```diff
@@ -12,6 +12,6 @@
 
     - name: "Download"
       become: yes
-      shell: "wget https://raw.githubusercontent.com/catchorg/Catch2/master/single_include/catch2/catch.hpp"
+      shell: "wget https://raw.githubusercontent.com/catchorg/Catch2/v2.x/single_include/catch2/catch.hpp"
```
csegarragonz, line 15: I needed this to compile the tests properly as the `master` branch was broken. It's a temporary workaround before #322 is merged. Shouldn't create too many conflicts.

tests/utils/CMakeLists.txt
```diff
@@ -12,11 +12,14 @@ set(LIB_FILES
 
 faasm_private_lib(test_utils "${LIB_FILES}")
 
+## TODO merge with upstreap (included as ext_package)
+find_package(Catch2 REQUIRED)
+
 target_link_libraries(test_utils
-        catch
+        Catch2::Catch2
```
csegarragonz, line 19: This is also part of the workaround, should be overwritten by #322.

src/wavm/mpi.cpp
```diff
@@ -429,6 +429,8 @@ WAVM_DEFINE_INTRINSIC_FUNCTION(env,
       "S - MPI_Probe {} {} {} {}", source, tag, comm, statusPtr);
     ContextWrapper ctx(comm);
     MPI_Status* status = &Runtime::memoryRef<MPI_Status>(ctx.memory, statusPtr);
+    // TODO Should Probe take the tag as a parameter as well?
```
csegarragonz, line 432: I was reluctant whether to include this message or not. After our discussion I have not payed attention on the performance of the chosen algorithm or it's complexity, but in this case I wanted to raise the question. Do we skip the message tag deliberately?

ansible/roles/openmpi/tasks/hostfile.yml
```diff
@@ -0,0 +1,6 @@
+---
```
Shillaker, line 1: This native MPI installation is only used for benchmarking our experiments, so isn't useful in "normal" faasm day-to-day work. The `openmpi` role already exists in the `faasm-experiments` repo (https://github.com/faasm/faasm-experiments/tree/master/ansible/roles/openmpi) so we should use that.

The general rule should be that all experimental code and benchmarks should live in `faasm-experiments`, and this repo should just contain the Faasm system.

It looks like an old `openmpi.yml` playbook was left in this repo by accident, so we should delete that.

func/mpi/mpi_onesided.cpp
```diff
@@ -44,9 +44,9 @@ FAASM_MAIN_FUNC()
     int* sharedData;
     MPI_Alloc_mem(memSize, MPI_INFO_NULL, &sharedData);
 
-    int putData[NUM_ELEMENT];
-    int expectedGetData[NUM_ELEMENT];
-    int expectedPutData[NUM_ELEMENT];
+    int* putData = new int[NUM_ELEMENT];
```
Shillaker, line 47: These should all be `auto` (i.e. avoid explicitly typing pointers that come from `new` like this)

func/mpi/mpi_wincreate.cpp
```diff
@@ -80,12 +80,17 @@ FAASM_MAIN_FUNC()
         }
 
         // Check size of window
-        checkIntAttr(window, MPI_WIN_SIZE, winSize, "MPI_WIN_SIZE");
-        checkIntAttr(window, MPI_WIN_DISP_UNIT, 1, "MPI_WIN_DISP_UNIT");
+        if (!checkIntAttr(window, MPI_WIN_SIZE, winSize, "MPI_WIN_SIZE"))
+            return 1;
+        if (!checkIntAttr(window, MPI_WIN_DISP_UNIT, 1, "MPI_WIN_DISP_UNIT"))
+            return 1;
```
Shillaker, line 86: We never do if statements without curly brackets, it's a consistent source of nasty bugs. 

I'm surprised this isn't captured in the code style

src/wavm/mpi.cpp
Shillaker: Not sure I understand. What are the pros and cons of each approach and which would you recommend?

It's better to leave TODOs as concrete actions if you can, rather than questions.

Shillaker (changes requested):

Shillaker (changes requested):
Looks good, just need to recompile the functions and check in the wasm (at the moment the tests in CI will still run with the old wasm)

csegarragonz:
Changes here are a bit out of place, but worth merging anyway. The only reason this wasn't still merged is that you suggested running `clang-tidy` and I went down that rabbit hole (never properly finishing it).

Shillaker (approved):

PR #327, opened 2020-10-21
Preserve user `host` and `port` in `inv invoke`

mkvoya:
The `host` doesn't take effect in `inv invoke demo hello -h 172.18.0.1` since it is overwritten by `get_invoke_host_port`.
This PR passes the user-provided `host` and `port` to `get_invoke_host_port`, so that they are not dropped when `knative` is true.


Shillaker (approved):
Great, LGTM!

Thanks for the PR @mkvoya

PR #332, opened 2020-10-26
Container tooling

csegarragonz:
In this PR I add some useful bits of scripting to ease development from within the container. I list them beneath.

**Switch to `docker-compose up` rather than `docker-compose run`:**
* According to Docker's [documentation](https://docs.docker.com/compose/faq/#whats-the-difference-between-up-run-and-start), it is preferrable to use `docker-compose up` rather than `docker-compose run`. In my opinion, it's the right tool if we want to manage _services_ rather than individual _containers_. However, it does come with some caveats.
* `docker-compose` has some freaky default behaviour with regard to where it looks for services. In our case (launching command from specific directory, not base, and using a custom compose file, not `docker-compose.yml`) this means that `docker-compose ps` outputs nothing even though servies are indeed running, and `docker exec service <command>` won't work either.
* The correct way to address the issue (albeit verbose) is to specify a `COMPOSE_PROJECT_NAME` (`faasm-dev`) in our case, and the compose file we have used.
* This means that when running any `compose` related command (`ps`, `logs`, `stop`, ...) we must prepend (after `docker-compose` and before the command) `-p faasm-dev -f docker/docker-compose-cli.yml`.

**Attach to a running `faasm/cli` instance rather than spinning a new one:**
* Before, running two times `bin/cli.sh` would yield two _different_ containers. Even though if most content is mounted to our local dir, it's preferrable to work on the _same_ container. This is useful if you are working on a set of files that are not necessarily mounted and persisted. It also makes more sense from a resource utilization standpoint. Moreover, given that compose services are `up`-ed using the dettached flag (`-d`), containers are persisted throughout invocations of `bin/cli.sh` transparently.
* With the changes included, the first time we invoke `bin/cli.sh` the services will be created and started (note the changes in the `cli` service definition to prevent it from exitting right away). Subsequent invocations will exec into the client container.

**Stopping services:**
* As previously introduced, launching compose in dettached mode means that the container will survive the user exitting it. This, if done with care, can be desirable, but could lead to unexpeced behaviour if overlooked. For instance changing the base image, not stopping the container explicitly, re-running `bin/cli.sh` to find out that nothing has changed.
* To manually stop all services, we can use compose's `stop` mechanism (perks of using `up` vs `run`), but with the previously specified prefix. I have added a line in the documentation explaining this.

**Another episode in the shell drama:**
* I have also solved the issue of `pip3` requirements not "persisting" in my custom image. In short, the problem is that `venv/bin/activate` was not being sourced.
* This was due to the, `bash`-specific (ehem heath-robinson), way of setting the `PROJ_ROOT` in `bin/workon.sh`. I have added a bit to the shell drama there. Fortunately this is only to be executed if not in container mode.
* If we are running within the container, I think it's easier to harcode the project root (and less error-prone, until we change it :rofl: ).


bin/cli.sh
```diff
@@ -12,12 +12,22 @@ VERSION=$(cat VERSION)
...
 INNER_SHELL=${SHELL:-"/bin/bash"}
 export CLI_IMAGE=${1:-${DEFAULT_IMAGE}}
+export COMPOSE_PROJECT_NAME="faasm-dev"
 echo "Running Faasm CLI (${CLI_IMAGE})"
 
-# Run shell in CLI container
+## Run shell in CLI container
+# If service not running, bring it up first.
+# Then attach to it (note we `up` containers in dettached mode).
+if [[ -z $(docker ps -q --filter name="faasm-dev_cli_1") ]]; then
```
Shillaker, line 21: Do we need this `if` line? Won't `docker-compose up` implicitly do this (i.e. it will ensure that the stuff is already running)?

Shillaker (commented):
Small comment re. `if`

docs/development.md
```diff
@@ -21,6 +21,12 @@ git submodule update --init
 ./bin/cli.sh
 ```
 
+The service orchestration is handled through `docker-compose`. If you want to
+stop all `faasm`-related development services run:
+```bash
+docker-compose -p faasm-dev -f docker/docker-compose-cli.yml stop
+```
+
```
Shillaker, line 29: Might be nice to put this in a script, e.g. `bin/stop-cli.sh`

bin/cli.sh
csegarragonz: The `if` line is not actually necessary. However, to ensure that containers are not recreated as we re-issue a `docker-compose up` command, we make sure we launch them using the `--no-recreate` flag.

Shillaker (approved):

PR #333, opened 2020-10-26
Incorporate Clang Tidy

csegarragonz:
The aim of this PR is to set up `clang-tidy` the LLVM-powered linter tool. Basic information about the project can be found in their [home page](https://clang.llvm.org/extra/clang-tidy/).

The tool is specially suited for `CMake`-based project (like `faasm`). It requires a compilation database in the form of a `compilation_commands.js` file. For future reference, this file can be generated running the following command:
```bash
cd <FAASM_REPO>
mkdir -p build; cd build
cmake --DCMAKE_EXPORT_COMPILE_COMMANDS=ON ..
``` 

The tool ships with a helper script `run-clang-tidy-10.py` a parallel `clang-tidy` runner to ease the analysis+fixing+re-format over a, sometimes large, set of files. In this PR I include a helper script to launch it.

There are a variety of checks the tool can perform. By default only the `clang-analyzer-*` static analysis checks are enabled. A [complete list](https://clang.llvm.org/extra/clang-tidy/checks/list.html) can be found online. In order to keep the scripts clean, I include the checks in a `.clang-tidy` config file. These are the checks that are currently included:
* [`readability-braces-around-statements`](https://clang.llvm.org/extra/clang-tidy/checks/readability-braces-around-statements.html): adds braces to single-line statements as demanded [here](https://github.com/faasm/faasm/pull/326#discussion_r509435836).
* [`modernize-use-auto`](https://clang.llvm.org/extra/clang-tidy/checks/modernize-use-auto.html): enforces the usage of the `auto` keyword as demanded [here](https://github.com/faasm/faasm/pull/326#discussion_r509435212).

These are the checks that (at least) I think should be included (**TBD!**):
* `modernize-*`: catch-all for enforcing modern (C++11) good practices.
* `readability-*`: I think most of them are worth it, but it's a matter of giving it a look.
I would personally include them all (i.e. `modernize-*,readability-*`,clang-analyzer-*`) and add others as we go.

Some other notes:
* For the time being the script does not actually change the code, just points to the mistakes. The `fix` flag needs to be enabled. Observe from the [checks list](https://clang.llvm.org/extra/clang-tidy/checks/list.html) that there's an additional column specifying whether that particular check includes automatic fixing or not.
* I have set the script to re-format the code after the changes (`-format`) according to a specific style. Passing `style 'file'` tells `clang-tidy` to look for a `.clang-format` file.
* There's also the option to embed `clang-tidy` as part of the [compilation toolchain](https://cmake.org/cmake/help/latest/prop_tgt/LANG_CLANG_TIDY.html). Remains to be discussed if this is something we are actually interested in or not.

**To-Discuss:**
* [x] Decide on a semi-definitive list of checks.
* [x] Add the `-fix` option (once discussion is done and before merging).
* [x] Include in the CI automation? Just for the modified files or all?
* [x] Include for other `faasm`-related repos (maybe symlinked like `.clang-format`)?
* [x] Include as part of the compilation toolchain?
* [x] Run only on `func` and `lib`.


Shillaker:
I think:

- First step, apply to the whole codebase (along with clang-format) once the SGX PR is merged in
- Run the check as part of CI (and fail if clang-tidy fails)
- Config looks good for now, let's iterate on this as we go
- Let's just copy the config file into Faabric and faasm-experiments once we're happy with it.

bin/run_clang_tidy.sh
```diff
@@ -0,0 +1,35 @@
...
+
+# Run clang-tidy on a set of files
+run-clang-tidy-10.py \
+    -config '' \
+    -j `nproc` \
+    -fix \
+    -format \
+    -style 'file' \
+    -p ${BUILD_PATH} \
+    -extra-arg-before='-I/build/faasm/_deps/faabric_ext-src/include/faabric/mpi' \
```
csegarragonz, line 30: Sometimes, `clang-tidy` will error out complaining it can't find the right `mpi` headers.

bin/run_clang_tidy.sh
```diff
@@ -0,0 +1,35 @@
...
+
+if [ -z "$1" ]; then
+    TARGET_DIR="."
+else
+    TARGET_DIR=$1
+fi
+
+pushd ${TARGET_DIR} >> /dev/null
+
+FILES=$(git ls-files {libs,func}/{"*.h","*.cpp","*.c"})
```
csegarragonz, line 18: As discussed offline, we only run the linting on `libs` and `func`.

.github/workflows/tests.yml
```diff
@@ -23,11 +23,12 @@ jobs:
       # --- Formatting checks ---
       - name: "Python formatting check"
         run: ./bin/check_black.sh
-      # Uncomment when all formatting done
       - name: "Run C/C++ formatting"
         run: ./bin/run_clang_format.sh
       - name: "Check C/C++ formatting changes"
         run: git diff --exit-code
+      - name: "Run C/C++ linter"
+        run: ./bin/run_clang_tidy.sh > /dev/null
```
csegarragonz, line 31: `clang-tidy` is _very_ verbose and difficult to read. This is because it inspects all header files that are compiled through the compilation database.

Discarding `stdout` leaves a leaner log, easier to inspect when something fails, as it just numbers the warnings generated and, if an error occurs, points to the file where it happened. Sufficient for CI imo.

ansible/roles/llvm10/tasks/main.yml
```diff
@@ -30,6 +30,7 @@
     name:
       - clang-10
       - clangd-10
+      - clang-tidy-10
```
csegarragonz, line 33: `clang-tidy-10` does not ship with `clang-tools-10` and it has to be installed separately. For the tests to pass then, we need to rebuild `cpp-root`, `base`, and `cli` containers.

.github/workflows/tests.yml
Shillaker: Yeah sounds good :+1:

ansible/roles/llvm10/tasks/main.yml
Shillaker: Is it really part of the base requirements for Faasm, or just the CLI container? I would say it's really only used in the CLI container, so should live in the Dockerfile for that. This file is too overblown as it is (e.g. we don't need `clang-tools` in the `worker` container)

.github/workflows/tests.yml
Shillaker: Yeah sounds good, should we put the `/dev/null` redirect in the script itself? Makes it nicer to run manually

bin/run_clang_tidy.sh
Shillaker: Why not `src` and `test`? Sorry I can't remember. Just because it spits out too many errors?

Shillaker (changes requested):

.github/workflows/tests.yml
csegarragonz: Indeed.

bin/run_clang_tidy.sh
csegarragonz: No. It's just for some reason I had noted down to run them only un `func` and `libs`. But it was something we said offline, discussing out loud, so no problem in adding these.

bin/run_clang_tidy.sh
```diff
@@ -0,0 +1,30 @@
...
+BUILD_PATH=/build/faasm/
+CONFIG=${FAASM_ROOT}/.clang-tidy
+
+FILES=$(git ls-files {func,libs,src,tests}/{"*.h","*.cpp","*.c"})
+
+if [ -z "$1" ]; then
+    :
+else
+    pushd $1 >> /dev/null
+    FILES=$(git ls-files {"*.h","*.cpp","*.c"})
```
csegarragonz, line 14: Before, running the `git ls-files` on a directory different than root would give no results (as no `func`, `libs`, ... would match).

Shillaker (approved):
Nice, can we gradually introduce more and more rules? This is great.

PR #337, opened 2020-11-05
Move final third-party deps to CMake

Shillaker:
WAMR and XRA are now loaded through CMake rather than Git submodules


csegarragonz (approved):
Nice! Way leaner now.

PR #338, opened 2020-11-09
Chaining with function pointers

Shillaker:
This changes the C/C++ interface to chain functions using function pointers rather than indexes.

Changes:

- Remove `FAASM_FUNC` directive (which was broken for WAVM execution anyway)
- Remove `FAASM_MAIN_FUNC` directive (switch to using standard `main()`)
- Change `faasm_chain_function` to `faasm_chain_name` (i.e. chaining a function by name)
- Change `faasm_chain_this` to `faasm_chain_ptr` (i.e. chaining a function with a function pointer)

Note that we still want to keep the chaining by name for use in other languages where passing pointers may not be possible. A lot of the WAVM-based implementation for this was already done as part of the distributed threading work.

The WAMR-based implementation will be part 2 of this work and done in a separate PR. The WAMR+SGX implementation will be part 3.


src/emulator/emulator.cpp
```diff
@@ -28,12 +28,8 @@ extern "C"
  */
 
 // Note thread-locality here
-static thread_local faabric::Message _emulatedCall = faabric::Message();
```
Shillaker, line 31: The emulator is almost useless now. It existed before Faabric and was used to execute native benchmarks with Faasm. Things have moved on a lot, so now it's just used for low-level state testing (which can eventually be moved to Faabric as well)

libs/cpp/faasm/pool.h
```diff
@@ -1,19 +0,0 @@
-#ifndef FAASM_POOL_H
-#define FAASM_POOL_H
```
Shillaker, line 2: This was never used

src/emulator/emulator.cpp
```diff
@@ -356,197 +350,42 @@ long __faasm_read_input(unsigned char* buffer, long bufferLen)
...
-    faabric::util::SystemConfig& conf = faabric::util::getSystemConfig();
-    if (conf.hostType == "knative") {
-        return _chain_knative(
-          _emulatedCall.function(), 0, name, buffer, bufferLen);
-    } else {
-        return _chain_local(0, name, buffer, bufferLen);
-    }
-}
-
-int _await_call_knative(unsigned int callId)
```
Shillaker, line 483: This stuff was all about supporting native code execution in Knative which we won't do any more.

faasmcli/faasmcli/tasks/dev.py
```diff
@@ -1,12 +1,19 @@
 from os import makedirs
 from os.path import exists
-from shutil import rmtree
 from subprocess import run
 
 from invoke import task
 
 from faasmcli.util.env import PROJ_ROOT, FAASM_BUILD_DIR
 
+DEV_TARGETS = [
```
Shillaker, line 9: Unrelated improvement for a shortcut for compiling all the development-related targets

faasmcli/faasmcli/tasks/run.py
```diff
@@ -1,23 +1,29 @@
+from copy import copy
 from invoke import task
+from os import environ
 
 from faasmcli.util.shell import run_command
 
 
 @task
-def user(ctx, user, repeats=1):
+def user(ctx, user, repeats=1, wamr=False):
```
Shillaker, line 9: Another slightly unrelated improvement to make it possible to pass a flag to run with WAMR

func/demo/CMakeLists.txt
```diff
@@ -100,9 +100,10 @@ demo_func(zygote_check zygote_check.cpp)
 if(FAASM_BUILD_TYPE STREQUAL "wasm")
     # demo_func(eigen_vec eigen_vec.cpp)
 
-    include_directories(/usr/local/faasm/llvm-sysroot/include/clapack)
-    demo_func_c(blas_check blas_check.c)
-    target_link_libraries(blas_check ${FAASM_BLAS_LIBS})
+    # TODO - fix BLAS check function
```
Shillaker, line 103: Fixing BLAS support will be a separate task

docs/cpp.md
```diff
@@ -17,66 +17,72 @@ You should then see the response `Hello faasm!`.
...
-FAASM_MAIN_FUNC() {
+int main(int argc, char* argv[]) {
     // Do something
 
     return 0;
 }
 ```
 
-Although for very simple functions, a standard `int main()` will also still work.
+Although for very simple functions, a standard `int main()` will also still
```
csegarragonz, line 37: Don't you think this is out of place without the macro?

csegarragonz (approved):

docs/cpp.md
Shillaker: Ah yes good point, I'll modify in part 2

PR #340, opened 2020-11-10
Make executable

csegarragonz:
One-liner. Forgot to add the `-x` flag.


Shillaker (approved):

Shillaker:
@csegarragonz could you try and reduce the number of commits in small PRs like this in future? Not sure why there are so many merge commits included in this one

PR #343, opened 2020-11-11
WAMR chaining with function pointers

Shillaker:
Update WAMR implementation to handle chaining with function pointers.

As mentioned in previous PR, this still doesn't include WAMR + SGX which I'll do in the next PR. Doing the SGX part properly requires writing a test with the simulator and a separate build, so it's worth putting in a separate PR.


src/wavm/syscalls.h
```diff
@@ -27,17 +27,6 @@ std::pair<std::string, std::string> getUserKeyPairFromWasm(int32_t keyPtr);
 
 std::string getMaskedPathFromWasm(int32_t strPtr);
 
-int awaitChainedCall(unsigned int messageId);
```
Shillaker, line 30: Moved this to somehwere non-WAVM-specific

tests/utils/matrix_utils.h
```diff
@@ -0,0 +1,12 @@
+#pragma once
```
Shillaker, line 1: Factored this out as importing Eigen with WAMR broke WAMR imports (they both seem to define a `Vector` class)

faasmcli/faasmcli/tasks/codegen.py
```diff
@@ -52,6 +52,7 @@ def local(ctx, wamr=False):
 
     # Always run the codegen required by the tests
     codegen(ctx, "demo", "echo", wamr=True)
+    codegen(ctx, "demo", "chain", wamr=True)
```
Shillaker, line 55: Eventually we'll run all the tests with both WAMR and WAVM, but for now we can just be tactical with WAMR

.github/workflows/tests.yml
```diff
@@ -49,7 +49,12 @@ jobs:
         run: ./bin/run_clang_format.sh
       - name: "Check C/C++ formatting changes"
         run: git diff --exit-code
-      # --- Compilation ---
+      # --- Code build ---
+      - name: "Rerun cmake"
+        run: inv -r faasmcli/faasmcli dev.cmake
+      - name: "Build dev tools"
+        run: inv -r faasmcli/faasmcli dev.tools
```
Shillaker, line 56: Better to use the same commands in the CLI and tests to build the code (it'll catch it if they break)

faasmcli/faasmcli/tasks/dev.py
```diff
@@ -35,7 +35,7 @@ def cmake(ctx, clean=False):
         PROJ_ROOT,
     ]
 
-    run(" ".join(cmd), shell=True, cwd=FAASM_BUILD_DIR)
+    run(" ".join(cmd), shell=True, check=True, cwd=FAASM_BUILD_DIR)
```
Shillaker, line 38: These weren't actually returning a non-zero error code if they failed

src/faaslet/Faaslet.cpp
```diff
@@ -87,48 +88,42 @@ void Faaslet::postBind(const faabric::Message& msg, bool force)
...
+                                          conf.wasmVm);
+        throw std::runtime_error("Unrecognised wasm VM");
     }
 }
 
 bool Faaslet::doExecute(faabric::Message& msg)
 {
     auto logger = faabric::util::getLogger();
 
-    // Force cold start if necessary
```
Shillaker, line 116: This was for simulating cold starts in experiments and is no longer necessary

csegarragonz (approved):

PR #344, opened 2020-11-12
Remove SGD code and smoke test

Shillaker:
This now lives in the experiments repo.

There's still a lot of generic SGD-related code that lives in the Faasm WebAssembly library. I'm going to wait a bit before deleting that as it's quite a nice test for complicated state interaction from WebAssembly.


tests/test/lib/test_sgd.cpp
```diff
@@ -231,105 +231,4 @@ TEST_CASE("Test reading errors from state", "[sgd]")
     double actual = faasm::readRootMeanSquaredError(p.nBatches, p.nTrain);
     REQUIRE(abs(actual - expectedRmse) < 0.0000001);
 }
-
-TEST_CASE("Run SGD smoke test", "[sgd]")
```
Shillaker, line 235: This was useful to spot errors when developing the experiment, but is no longer necessary

csegarragonz (approved):

PR #351, opened 2020-11-13
MPI stubs and MPI_Cart family implementation

csegarragonz:
In this PR I include the stub definitions for the methods included in faasm/faabric#20, which in turn are necessary to run LAMMPS with MPI. Most of the changes are implementation-less.

The only bit of code I have added is the one corresponding to the `MPI_Cart*` family of methods. They are responsible to, given a standard mpi communicator (of type `MPI_Comm`), return a new one with an attached cartesian topology (namely, processes are orgnaized in a multi-dim grid).

We, mostly, ignore the parameters passed by the user, and by default return a 2-dim grid with as many processes as we can fit (given the current mpi world). Note that this is a safer approach than `noop`-ing, as the return values may eventually be used.


src/wavm/mpi.cpp
```diff
@@ -917,7 +1473,7 @@ WAVM_DEFINE_INTRINSIC_FUNCTION(env,
                                "MPI_Get",
                                I32,
                                MPI_Get,
-                               I32 recvBuff,
+                               I32 recvBuf,
```
csegarragonz, line 1476: We use Buf with a single `f` elsewhere.

src/wavm/mpi.cpp
```diff
@@ -168,6 +169,26 @@ WAVM_DEFINE_INTRINSIC_FUNCTION(env, "MPI_Init", I32, MPI_Init, I32 a, I32 b)
...
+ * TODO not implemented.
+ */
+WAVM_DEFINE_INTRINSIC_FUNCTION(env,
+                               "MPI_Get_version",
+                               I32,
+                               MPI_Get_version,
+                               I32 version,
+                               I32 subversion)
+{
+    faabric::util::getLogger()->debug("S - MPI_Get_version");
```
Shillaker, line 185: For debugging purposes I personally like to put the arguments in these logging statements. Not worth changing now, can just do in another PR.

src/wavm/mpi.cpp
```diff
@@ -233,6 +349,44 @@ WAVM_DEFINE_INTRINSIC_FUNCTION(env,
...
+                               I32,
+                               MPI_Rsend,
+                               I32 buffer,
+                               I32 count,
+                               I32 datatype,
+                               I32 destRank,
+                               I32 tag,
+                               I32 comm)
+{
+    faabric::util::getLogger()->debug("S - MPI_Rsend {} {} {} {} {} {} {}",
```
Shillaker, line 367: Too many `{}`s in this string

src/wavm/mpi.cpp
```diff
@@ -399,6 +585,47 @@ WAVM_DEFINE_INTRINSIC_FUNCTION(env,
...
+                               "MPI_Waitany",
+                               I32,
+                               MPI_Waitany,
+                               I32 count,
+                               I32 requestArray,
+                               I32 index,
+                               I32 status)
+{
+    faabric::util::getLogger()->debug(
+      "S - MPI_Waitany {} {} {} {}", count, requestArray, index, status);
```
Shillaker, line 622: Is `index` a keyword for something? It's highlighted in this diff but I can't tell. Maybe use `idx` to be safe

src/wavm/mpi.cpp
```diff
@@ -449,7 +676,7 @@ WAVM_DEFINE_INTRINSIC_FUNCTION(env,
                                I32 comm)
 {
     faabric::util::getLogger()->debug(
-      "S - MPI_Bcast {} {} {} {}", buffer, count, datatype, root, comm);
+      "S - MPI_Bcast {} {} {} {} {}", buffer, count, datatype, root, comm);
```
Shillaker, line 679: Good spot

src/wavm/mpi.cpp
```diff
@@ -843,6 +1202,202 @@ WAVM_DEFINE_INTRINSIC_FUNCTION(env,
...
+      "S - MPI_Cart_rank {} {} {}", comm, coords, rankPtr);
+
+    ContextWrapper ctx(comm);
+
+    int* coordsArray = Runtime::memoryArrayPtr<int>(
+      ctx.memory, (Uptr)coords, MPI_CART_MAX_DIMENSIONS);
+    int sideLength =
+      static_cast<int>(std::floor(std::sqrt(ctx.world.getSize())));
+    int rank = sideLength * coordsArray[0] + coordsArray[1];
+    ctx.writeMpiResult<int>(rankPtr, rank);
```
Shillaker, line 1270: This logic should be moved into the world itself and have a small test. The function might be `ctx.world.getCartesianRank()`

src/wavm/mpi.cpp
```diff
@@ -843,6 +1202,202 @@ WAVM_DEFINE_INTRINSIC_FUNCTION(env,
...
+{
+    faabric::util::getLogger()->debug(
+      "S - MPI_Cart_get {} {} {} {} {}", comm, maxdims, dims, periods, coords);
+
+    ContextWrapper ctx(comm);
+
+    auto commPtr =
+      &Runtime::memoryRef<faabric_communicator_t>(ctx.memory, comm);
+    faabric::util::getLogger()->debug(
+      "Cart_get w/ comm: {}. Id: {}", comm, commPtr->id);
```
Shillaker, line 1303: This sort of print debugging should be removed (there should be a test to check this works)

src/wavm/mpi.cpp
```diff
@@ -843,6 +1202,202 @@ WAVM_DEFINE_INTRINSIC_FUNCTION(env,
...
+    int sideLength =
+      static_cast<int>(std::floor(std::sqrt(ctx.world.getSize())));
+    int nprocs = sideLength * sideLength;
+    for (uint8_t i = 0; i < MPI_CART_MAX_DIMENSIONS; i++) {
+        nprocs /= sideLength;
+        dimsArray[i] = sideLength;
+        periodsArray[i] = 0;
+        coordsArray[i] = ctx.rank > nprocs ? MPI_UNDEFINED : rank / nprocs;
+        rank %= nprocs;
+    }
```
Shillaker, line 1335: This logic should be moved to the world itself and tested

src/wavm/mpi.cpp
```diff
@@ -1103,6 +1675,23 @@ WAVM_DEFINE_INTRINSIC_FUNCTION(env,
     return MPI_SUCCESS;
 }
 
+/** Frees a data type
```
Shillaker, line 1678: Inconsistent style on these comments

Shillaker (changes requested):
Looks good, but all non-wasm related logic should be moved into the world itself and have an associated unit test. Should also have a test function in `func/mpi` as with other MPI operations.

src/wavm/mpi.cpp
csegarragonz: Godd catch

src/wavm/mpi.cpp
csegarragonz: I totally missed this one. Thanks

csegarragonz:
This should be merged in after faasm/faabric#22 as it depends on the implementation there (otherwise the tests won't pass).

func/mpi/mpi_cartesian.cpp
```diff
@@ -0,0 +1,76 @@
...
+    }
+
+    // Test integrity of results
+    if (!faasm::compareArrays<int>(coords, expected, 2)) {
+        printf("Wrong cartesian coordinates.\n");
+        return 1;
+    }
+    printf("MPI_Cart_get succesful check.\n");
+
+    printf("still here\n");
```
Shillaker, line 54: Could probably remove this

src/wavm/mpi.cpp
```diff
@@ -333,6 +488,48 @@ WAVM_DEFINE_INTRINSIC_FUNCTION(env,
...
+      recvCount,
+      recvType,
+      source,
+      recvTag,
+      comm,
+      statusPtr);
+
+    throw std::runtime_error("MPI_Sendrecv not implemented!");
+
+    return 0;
```
Shillaker, line 530: Should this be `return MPI_SUCCESS` to be in line with the rest of the file?

tests/test/faaslet/test_mpi.cpp
```diff
@@ -36,6 +36,11 @@ TEST_CASE("Test MPI broadcast", "[wasm]")
     checkMpiFunc("mpi_bcast");
 }
 
+TEST_CASE("Test MPI cartesian coordinates", "[wasm]")
```
Shillaker, line 39: There are a lot of tests now tagged with `wasm`, perhaps in future we could tag them with something more meaningful or with multiple tags (e.g. tagging these ones with `mpi` so we could do `/build/faasm/bin/tests "[mpi]"`). Not something to worry about now though

Shillaker (commented):
Nice, LGTM. A couple of optional tweaks in the comments.

tests/test/faaslet/test_mpi.cpp
csegarragonz: Will handle this quickly now. I also thought about it.

Shillaker (approved):

csegarragonz:
This PR is blocked as it needs #360 to pass the tests.

PR #352, opened 2020-11-13
Create container for native tools

Shillaker:
These tools are used for building external projects, most notably the pyfaasm repo and tests


Shillaker:
This is no longer needed, if we move all the C/C++-specific code to faasm-cpp

PR #355, opened 2020-11-16
Towards a fix for Python build issues

Shillaker:
There are several improvements in here to catch issues with the Python-related testing, but unfortunately it's still not fixed.

I'd like to move onto something else but make use of these changes, so would be good to merge.


.github/workflows/tests.yml
```diff
@@ -70,12 +70,10 @@ jobs:
         run: inv -r faasmcli/faasmcli libs.fake
       - name: "Run codegen"
         run: inv -r faasmcli/faasmcli codegen.local
-      - name: "Run WAMR codegen"
```
Shillaker, line 73: `codegen.local` does the WAMR codegen

.github/workflows/tests.yml
```diff
@@ -70,12 +70,10 @@ jobs:
...
-        run: inv -r faasmcli/faasmcli codegen demo echo --wamr
+      - name: "Run python codegen"
+        run: inv -r faasmcli/faasmcli python.codegen
       - name: "Copy Python functions into place"
         run: inv -r faasmcli/faasmcli upload.user python --py --local-copy
-      - name: "Make sure permissions are properly set"
-        run: chown -R root:root /usr/local/faasm
```
Shillaker, line 78: This is no longer necessary in the containerised world

.github/workflows/tests.yml
```diff
@@ -70,12 +70,10 @@ jobs:
         run: inv -r faasmcli/faasmcli libs.fake
       - name: "Run codegen"
         run: inv -r faasmcli/faasmcli codegen.local
-      - name: "Run WAMR codegen"
-        run: inv -r faasmcli/faasmcli codegen demo echo --wamr
+      - name: "Run python codegen"
+        run: inv -r faasmcli/faasmcli python.codegen
```
Shillaker, line 74: This step in the `codegen.local` task was failing so I've added it explicitly here

func/python/py_func.cpp
```diff
@@ -161,13 +164,21 @@ int main(int argc, char* argv[])
...
             if (returnValue != nullptr) {
-                printf("Python call succeeded (return value = %ld)\n",
-                       PyLong_AsLong(returnValue));
+                long r = PyLong_AsLong(returnValue);
+
+                if (r == 0) {
+                    printf("Python call succeeded\n");
+                } else {
+                    printf("Python call failed (return value = %ld)\n", r);
+                }
```
Shillaker, line 173: This now more strictly enforces Python functions returning zero to be successful (rather than just returning anything)

.github/workflows/tests.yml
```diff
@@ -41,6 +41,12 @@ jobs:
         run: git checkout --force ci-branch
       - name: "Log commits"
         run: git log -3
+      # --- Cache ---
+      - name: "Configure build cache"
+        uses: actions/cache@v2
+        with:
+          path: /usr/local/faasm/object
+          key: ${{ runner.os }}-cache
```
Shillaker, line 49: Putting this here was a bit tactical to avoid merge conflicts with #356

faasmcli/faasmcli/tasks/codegen.py
```diff
@@ -54,7 +54,3 @@ def local(ctx, wamr=False):
     # Always run the codegen required by the tests
     codegen(ctx, "demo", "echo", wamr=True)
     codegen(ctx, "demo", "chain", wamr=True)
-
-    print("Running codegen on python shared objects")
-    binary = find_codegen_shared_lib()
-    check_output("{} {}".format(binary, PY_RUNTIME_ROOT), shell=True)
```
Shillaker, line 60: Done separately with an explicit command now

csegarragonz (approved):

PR #356, opened 2020-11-17
Codegen hashing

Shillaker:
Hash the inputs to all codegen, and skip if the hash hasn't changed.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -41,6 +41,12 @@ jobs:
         run: git checkout --force ci-branch
       - name: "Log commits"
         run: git log -3
+      # --- Cache ---
+      - name: "Configure build cache"
+        uses: actions/cache@v2
+        with:
+          path: /usr/local/faasm/object
+          key: ${{ runner.os }}-cache
       # --- Formatting checks ---
       - name: "Python formatting check"
         run: ./bin/check_black.sh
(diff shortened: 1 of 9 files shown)


.github/workflows/tests.yml
```diff
@@ -41,6 +41,12 @@ jobs:
         run: git checkout --force ci-branch
       - name: "Log commits"
         run: git log -3
+      # --- Cache ---
+      - name: "Configure build cache"
+        uses: actions/cache@v2
+        with:
+          path: /usr/local/faasm/object
+          key: ${{ runner.os }}-cache
```
Shillaker, line 49: We'll see how consistent the architecture of the Github runners is. This may have to become specific to the underlying processor

include/storage/FileserverFileLoader.h
```diff
@@ -30,6 +30,28 @@ class FileserverFileLoader : public FileLoader
 
     std::vector<uint8_t> loadSharedFile(const std::string& path) override;
 
+    std::vector<uint8_t> loadFunctionObjectHash(
```
Shillaker, line 33: Using pure virtual functions like this is a bit nasty as you have to override them for every single subclass, but I didn't want to go down another rabbit hole refactoring the `FileLoader` class.

csegarragonz (approved):

PR #357, opened 2020-11-19
SGX Attestation (WiP)

dgoltzsche:
This adds the possibility of SGX attestation. Credits go to Lennard Golsch and @J-Heinemann

PR #358, opened 2020-11-19
SGX tests in CI

Shillaker:


faasmcli/faasmcli/tasks/codegen.py
```diff
@@ -1,18 +1,19 @@
...
 
 
 @task(default=True)
 def codegen(ctx, user, function, wamr=False):
     """
     Generates machine code for the given function
     """
-    env = {"WASM_VM": "wamr" if wamr else "wavm"}
+    env = copy(environ)
+    env.update({"WASM_VM": "wamr" if wamr else "wavm"})
```
Shillaker, line 16: This wasn't picking up the log level set in the external environment

src/storage/FileLoader.cpp
```diff
@@ -76,6 +76,10 @@ void FileLoader::codegenForFunction(faabric::Message& msg)
     if ((!oldHash.empty()) && newHash == oldHash) {
         logger->debug("Skipping codegen for {}", funcStr);
         return;
+    } else if (oldHash.empty()) {
+        logger->debug("No old hash found for {}", funcStr);
+    } else {
+        logger->debug("Hashes differ for {}", funcStr);
```
Shillaker, line 82: This is unrelated but useful for debugging

tests/test/sgx/test_sgx.cpp
```diff
@@ -0,0 +1,16 @@
...
+#include <faabric/util/func.h>
+#include <wamr/WAMRWasmModule.h>
+
+using namespace wasm;
+
+namespace tests {
+TEST_CASE("Test executing echo function with SGX", "[sgx]")
+{
+    // TODO - uncomment this to actually run the SGX test
+    // executeWithSGX("demo", "chain");
```
Shillaker, line 14: This currently fails but should be fixed in #357 , so I'll leave it for @dgoltzsche to uncomment.

csegarragonz (approved):

PR #359, opened 2020-11-19
Fix directory listing bug

Shillaker:
This addresses a hard-to-diagnose issue which was causing certain Python functions to fail. The bug manifested itself as Python being unable to import `weakref` on some platforms but not others. It came down to an issue where the `__wasi_fd_readdir` implementation would drop one or two directory entries on larger directory listings. Which directories got dropped depended on the ordering of the native `readdir` syscall, hence the bug was platform dependent.

On certain platforms, the `weakref.py` file was dropped from the listing of the `lib/python3.8` dir, hence Python was claiming that the `weakref` module didn't exist. On other platforms, non-essential entries for this dir were dropped, hence Python executed successfully.

This PR restructures the `__wasi_fd_readdir` call to:

- Prefetch the full directory listing behind the scenes (this avoids having two iterators going at once)
- Completely encapsulate the logic for `__wasi_fd_readdir` in the `FileDescriptor` class rather than have it split across two files
- Add more tests for the directory iterator and buffer handling

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -51,8 +51,9 @@ jobs:
         ports:
           - 6379:6379
     steps:
-      - name: "ulimits"
-        run: ulimit -a
+      # --- Environment ---
+      - name: "Directory permissions"
+        run: chown -R root:root /usr/local/faasm
       # --- Check out code ---
       - name: "Fetch ref"
         run: git fetch origin ${GITHUB_REF}:ci-branch
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -88,5 +88,7 @@ jobs:
       - name: "Copy Python functions into place"
         run: inv -r faasmcli/faasmcli upload.user python --py --local-copy
       # --- Test run ---
+      - name: "Just run a the numpy test"
+        run: inv -r faasmcli/faasmcli run python numpy_test
       - name: "Run the tests"
         run: /build/faasm/bin/tests
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -51,13 +51,6 @@ jobs:
         ports:
           - 6379:6379
     steps:
-      # --- Environment ---
-      - name: "CPU info"
-        run: cat /proc/cpuinfo
-      - name: "Get CPU model name"
-        run: echo "CPU_MODEL=$(./bin/print_cpu.sh)" >> $GITHUB_ENV
-      - name: "Print CPU model"
-        run: echo ${{ env.CPU_MODEL}}
       # --- Check out code ---
       - name: "Fetch ref"
         run: git fetch origin ${GITHUB_REF}:ci-branch
@@ -65,6 +58,13 @@ jobs:
         run: git checkout --force ci-branch
       - name: "Log commits"
         run: git log -3
+      # --- CPU information ---
+      - name: "CPU info"
+        run: cat /proc/cpuinfo
+      - name: "Get CPU model name"
+        run: echo "CPU_MODEL=$(./bin/print_cpu.sh)" >> $GITHUB_ENV
+      - name: "Print CPU model"
+        run: echo ${{ env.CPU_MODEL}}
       # --- Cache ---
       - name: "Configure machine code cache"
         uses: actions/cache@v2
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -28,6 +28,8 @@ jobs:
         run: ./bin/run_clang_format.sh
       - name: "Check C/C++ formatting changes"
         run: git diff --exit-code
+      - name: "Check weakref is actuall there"
+        run: "ls -al /usr/local/faasm/runtime_root/lib/python3.8/weakref.py"
 
   build:
     if: github.event.pull_request.draft == false
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -51,9 +51,6 @@ jobs:
         ports:
           - 6379:6379
     steps:
-      # --- Environment ---
-      - name: "Directory permissions"
-        run: chown -R root:root /usr/local/faasm
       # --- Check out code ---
       - name: "Fetch ref"
         run: git fetch origin ${GITHUB_REF}:ci-branch
@@ -93,11 +90,5 @@ jobs:
       - name: "Clear existing pyc files"
         run: inv -r faasmcli/faasmcli python.clear-runtime-pyc
       # --- Test run ---
-      - name: "Run the listdir check"
-        run: inv -r faasmcli/faasmcli run python dir_test
-      - name: "Run the Python language check"
-        run: inv -r faasmcli/faasmcli run python lang_test
-      - name: "Run the the numpy check"
-        run: inv -r faasmcli/faasmcli run python numpy_test
       - name: "Run the tests"
         run: /build/faasm/bin/tests
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -91,7 +91,9 @@ jobs:
       - name: "Copy Python functions into place"
         run: inv -r faasmcli/faasmcli upload.user python --py --local-copy
       # --- Test run ---
-      - name: "Just run a the numpy test"
+      - name: "Run the Python language check"
+        run: inv -r faasmcli/faasmcli run python lang_test
+      - name: "Run the the numpy check"
         run: inv -r faasmcli/faasmcli run python numpy_test
       - name: "Run the tests"
         run: /build/faasm/bin/tests
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -64,7 +64,7 @@ jobs:
       - name: "Get CPU model name"
         run: echo "CPU_MODEL=$(./bin/print_cpu.sh)" >> $GITHUB_ENV
       - name: "Print CPU model"
-        run: echo ${{ env.CPU_MODEL}}
+        run: echo "${{ env.CPU_MODEL}}"
       # --- Cache ---
       - name: "Configure machine code cache"
         uses: actions/cache@v2
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -51,6 +51,8 @@ jobs:
         ports:
           - 6379:6379
     steps:
+      - name: "ulimits"
+        run: ulimit -a
       # --- Check out code ---
       - name: "Fetch ref"
         run: git fetch origin ${GITHUB_REF}:ci-branch
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -34,12 +34,12 @@ jobs:
     runs-on: ubuntu-20.04
     env:
       CGROUP_MODE: off
-      NETNS_MODE: off
       HOST_TYPE: ci
-      REDIS_STATE_HOST: redis
-      REDIS_QUEUE_HOST: redis
       LOG_LEVEL: info
       MAX_FAASLETS: 5
+      NETNS_MODE: off
+      REDIS_QUEUE_HOST: redis
+      REDIS_STATE_HOST: redis
     container:
       image: faasm/cli:0.5.3
     defaults:
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -54,10 +54,10 @@ jobs:
       # --- Environment ---
       - name: "CPU info"
         run: cat /proc/cpuinfo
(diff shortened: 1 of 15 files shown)


include/storage/FileDescriptor.h
```diff
@@ -37,9 +41,9 @@ ReadWriteType getRwType(uint64_t rights);
 class DirEnt
 {
   public:
-    unsigned int next;
+    uint64_t next;
     uint8_t type;
-    unsigned int ino;
+    uint64_t ino;
```
Shillaker, line 46: Being explicit here to make sure it's easy to compare with the WASI spec

func/python/dir_test.py
```diff
@@ -1,23 +1,23 @@
...
-        print("Dirlist fine = {}".format(dirlist))
+    dirlist = os.listdir(path)
+    dirlist.sort()
+    dir_list_str = ",".join(dirlist)
 
-        for file in dirlist:
-            print(file)
+    print(dir_list_str)
+
+    set_output(bytes(dir_list_str, "utf-8"))
```
Shillaker, line 21: This whole file is now used in a test to check Python's own listdir function (which was the source of all the issues)

src/storage/FileDescriptor.cpp
```diff
@@ -178,76 +180,180 @@ FileDescriptor FileDescriptor::stderrFactory()
...
-  : iterStarted(false)
-  , iterFinished(false)
-  , dirPtr(nullptr)
-  , rightsSet(false)
-  , linuxFd(-1)
-  , linuxMode(-1)
-  , linuxFlags(-1)
-  , linuxErrno(0)
-  , wasiErrno(0)
-{}
```
Shillaker, line 191: These are now set in the class definition

src/wavm/io.cpp
```diff
@@ -183,59 +183,25 @@ WAVM_DEFINE_INTRINSIC_FUNCTION(wasi,
...
-            // Copy its name in straight after
-            int pathCopySize =
-              faabric::util::safeCopyToBuffer(BYTES_CONST(dirEnt.path.c_str()),
-                                              dirEnt.path.size(),
-                                              buffer + bytesCopied,
-                                              bytesLeft);
-            bytesCopied += pathCopySize;
-            bytesLeft -= pathCopySize;
-        }
-    }
```
Shillaker, line 238: This logic now lives in the `FileDescriptor` class

tests/test/faaslet/test_python.cpp
```diff
@@ -68,26 +121,18 @@ TEST_CASE("Test python state write/ read", "[faaslet]")
     writeCall.set_pythonuser("python");
     writeCall.set_pythonfunction("state_test_write");
     writeCall.set_ispython(true);
-    faaslet::Faaslet faaslet = execFunction(writeCall);
+    execFunction(writeCall);
```
Shillaker, line 124: I've simplified the `execFunction` interface as it was adding unnecessary complexity to just execute a single function and check it doesn't fail

tests/test/faaslet/test_python.cpp
```diff
@@ -58,7 +111,7 @@ TEST_CASE("Test python echo", "[faaslet]")
     REQUIRE(result == input);
 }
 
-TEST_CASE("Test python state write/ read", "[faaslet]")
+TEST_CASE("Test python state write/ read", "[python]")
```
Shillaker, line 114: `[faaslet]` tag is overloaded, `[python]` is more useful to group similar tests

tests/test/wasm/test_cloning.cpp
```diff
@@ -231,7 +231,7 @@ TEST_CASE("Test cloned execution on simple module", "[wasm]")
     }
 }
 
-TEST_CASE("Test cloned execution on complex module", "[wasm][!mayfail]")
+TEST_CASE("Test cloned execution on complex module", "[wasm]")
```
Shillaker, line 234: We don't expect these tests to fail any more

tests/utils/worker_utils.cpp
```diff
@@ -7,56 +7,32 @@
...
-
-    // Now execute the function
-    w.processNextMessage();
-    REQUIRE(sch.getFunctionInFlightCount(call) == 0);
-    REQUIRE(sch.getFunctionWarmNodeCount(call) == 1);
-    REQUIRE(bindQueue->size() == 0);
-
-    // Check success
-    faabric::Message result = sch.getFunctionResult(call.id(), 1);
-    REQUIRE(result.returnvalue() == 0);
```
Shillaker, line 51: Mentioned above, excessive complexity around executing a single function in a test

src/storage/FileDescriptor.cpp
```diff
@@ -178,76 +180,180 @@ FileDescriptor FileDescriptor::stderrFactory()
...
-    dirPtr = nullptr;
-    iterStarted = false;
-    iterFinished = false;
+    dirContentsLoaded = false;
+    dirContentsIdx = 0;
+    dirContents.clear();
 }
 
-DirEnt FileDescriptor::iterNext()
+void FileDescriptor::loadDirContents()
```
Shillaker, line 191: The diff here is quite hard to read. I've basically split this up into a function which prefetches the whole directory (`loadDirContents`), and a much simpler one that handles the iteration (`iterNext`)

src/storage/FileDescriptor.cpp
```diff
@@ -178,76 +180,180 @@ FileDescriptor FileDescriptor::stderrFactory()
...
+ * The external libc wrapper should prevent this third point.
+ *
+ * If the final dirent or path doesn't fit, the caller will expect the next
+ * result to contain that same dirent/ path, so we have to step back one in the
+ * iterator if that's the case.
+ *
+ * The caller knows they've reached the end when the buffer is _not_ filled by
+ * this call.
+ */
+size_t FileDescriptor::copyDirentsToWasiBuffer(uint8_t* buffer,
```
Shillaker, line 305: The logic here is the stuff we need to get right for everything else to work. It used to be held in `src/wavm/io.cpp` but it's now captured here and tested as part of this class.

tests/test/faaslet/test_python.cpp
```diff
@@ -19,17 +21,68 @@ void checkPythonFunction(const std::string& funcName)
     execFunction(call);
 }
 
-TEST_CASE("Test python conformance", "[faaslet]")
+TEST_CASE("Test Python listdir", "[python]")
+{
```
Shillaker, line 25: This is the missing test that would have caught the issue

tests/test/module_cache/test_module_cache.cpp
```diff
@@ -34,8 +34,12 @@ TEST_CASE("Test creating zygotes", "[zygote]")
     REQUIRE(moduleA.isBound());
 
     // Execute the function normally and make sure zygote is not used directly
-    faaslet::Faaslet faaslet = execFunction(msgA);
-    REQUIRE(faaslet.isBound());
+    faaslet::Faaslet faaslet(0);
+    faaslet.bindToFunction(msgA);
+    std::string errorMessage = faaslet.executeCall(msgA);
+    REQUIRE(errorMessage.empty());
+    REQUIRE(msgA.returnvalue() == 0);
```
Shillaker, line 41: Just refactoring to fit with the simpler testing interface

tests/test/storage/test_file_descriptor.cpp
```diff
@@ -10,18 +11,12 @@
 using namespace storage;
 
 namespace tests {
-FileDescriptor& getRootFd()
+
+TEST_CASE("Test stat and mkdir", "[storage]")
```
Shillaker, line 15: All of the tests in this file were called `"Check..."` but `"Test..."` is more common in the codebase.

tests/test/storage/test_file_descriptor.cpp
```diff
@@ -10,18 +11,12 @@
...
 {
     FileSystem fs;
     fs.prepareFilesystem();
-
-    FileDescriptor& rootFd = fs.getFileDescriptor(4);
-    return rootFd;
```
Shillaker, line 19: I've now avoided hardcoding this number `4` everywhere

.github/workflows/tests.yml
```diff
@@ -34,12 +34,12 @@ jobs:
...
       CGROUP_MODE: off
-      NETNS_MODE: off
       HOST_TYPE: ci
-      REDIS_STATE_HOST: redis
-      REDIS_QUEUE_HOST: redis
       LOG_LEVEL: info
       MAX_FAASLETS: 5
+      NETNS_MODE: off
+      REDIS_QUEUE_HOST: redis
+      REDIS_STATE_HOST: redis
```
Shillaker, line 42: Alphabetical

.github/workflows/tests.yml
```diff
@@ -81,14 +81,14 @@ jobs:
...
-      - name: "Set up test lib"
-        run: inv -r faasmcli/faasmcli libs.fake
       - name: "Run codegen"
         run: inv -r faasmcli/faasmcli codegen.local
       - name: "Run python codegen"
         run: inv -r faasmcli/faasmcli python.codegen
       - name: "Copy Python functions into place"
         run: inv -r faasmcli/faasmcli upload.user python --py --local-copy
+      - name: "Clear existing pyc files"
+        run: inv -r faasmcli/faasmcli python.clear-runtime-pyc
```
Shillaker, line 91: This is a safety step that was missing before, in case pyc files are left over from the cpython container build

tests/test/storage/test_file_descriptor.cpp
```diff
@@ -238,4 +227,121 @@ TEST_CASE("Check stat and read shared file", "[storage]")
...
+        std::vector<uint8_t> buffer(sizeA + sizeB + sizeC - 10);
+
+        // Copy into this buffer
+        size_t bytesCopied =
+          fileDesc.copyDirentsToWasiBuffer(buffer.data(), buffer.size());
+        REQUIRE(bytesCopied == buffer.size());
+
+        // Check contents
+        checkWasiDirentInBuffer(buffer.data(), entA);
+        checkWasiDirentInBuffer(buffer.data() + sizeA, entB);
```
csegarragonz, line 335: Could we actually check that `entC` is _not_ in the buffer?

csegarragonz (changes requested):
Other than the suggested changes LGTM! Glad you sorted this.

src/storage/FileDescriptor.cpp
Shillaker: Yep, this is what the libc wrapper expects

src/storage/FileDescriptor.cpp
```diff
@@ -178,76 +180,180 @@ FileDescriptor FileDescriptor::stderrFactory()
...
+        auto pathPtr = BYTES_CONST(d.path.c_str());
+        std::copy(pathPtr, pathPtr + pathSize, buffer);
+        bytesLeft -= pathSize;
+        buffer += pathSize;
+    }
+
+    if (isBufferFull) {
+        // Go back one, we've not been able to finish this entry
+        iterBack();
+        return bufferLen;
```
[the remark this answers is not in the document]
Shillaker: The contract with the libc wrapper is:

- If buffer is full = you've not finished yet
- If buffer is not full = you've finished

So even when the buffer is not full, but there's more to come, you need to _say_ the buffer is full.

tests/test/storage/test_file_descriptor.cpp
Shillaker: Only part of `entC` would fit, so we'd be checking whether _some_ of `entC` had been written to the buffer (might be a bit fiddly). 

More importantly, it's undefined as to whether the function should or shouldn't write this partial dirent. The libc wrapper will always skip reading this last part of the buffer, unless there's at least a full dirent's worth of bytes left (i.e. it will never read this part of the buffer and it's undefined as to what should be written there).

src/storage/FileDescriptor.cpp
```diff
@@ -178,76 +180,180 @@ FileDescriptor FileDescriptor::stderrFactory()
...
+        bytesLeft -= wasiDirentSize;
+        buffer += wasiDirentSize;
+
+        // Drop out if we have no room for the path
+        if (bytesLeft < pathSize) {
+            isBufferFull = true;
+            break;
+        }
+
+        auto pathPtr = BYTES_CONST(d.path.c_str());
```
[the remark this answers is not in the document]
Shillaker: Yes, because `d.path.c_str()` is a const pointer, whereas the other one isn't (and you can't cast a const pointer to a non-const pointer)

csegarragonz (approved):

PR #360, opened 2020-11-23
Bump Faabric version to 0.0.13

csegarragonz:


.github/workflows/tests.yml
```diff
@@ -66,7 +66,7 @@ jobs:
           key: ${{ runner.os }}-cache
       # --- Code build ---
       - name: "Build dev tools"
-        run: inv -r faasmcli/faasmcli dev.tools --build Release
+        run: inv -r faasmcli/faasmcli dev.tools --clean --build Release
```
csegarragonz, line 69: This was the reason why the new tag wasn't being properly fetched from Git. It's still not clear to me what exactly is cached or not during this CI builds. The most I can say is that, empirically, adding this flag forces the fetch (otherwise only `CMake`' s populate step was triggered).

csegarragonz:
Now this fails where expected (not fetching contents). I am happy to merge this after #359

.github/workflows/tests.yml
Shillaker: AFACIT the issue is that CMake isn't detecting that the tag has changed on the external Faabric dependency, therefore we need to rerun the CMake configuration. Have you tried adding the following?

```
- name: "Rerun cmake config"
  run: inv -r faasmcli/faasmcli dev.cmake --build Release
```

The containers are structured so that we do a CMake build in the base container, then _rebuild_ it in the tests (rather than building from scratch). This is to avoid constantly downloading and rebuilding the dependencies.

The change here means that we completely nuke the existing CMake build, and build everything from scratch. So while it does work in picking up the new tag, it might add a couple of minutes to the build.

Shillaker (changes requested):

.github/workflows/tests.yml
csegarragonz: `inv -r faasmcli/faasmcli dev.tools --build Release` already calls `cmake`under the hood, so adding those two lines won't make a difference.

I see your point, maybe we can discuss this in a bit today (in the meantime I'll do some reading on cmake).

Shillaker (commented):
Provided the tests keep passing this looks good :+1:

.github/workflows/tests.yml
Shillaker: Ah yes sorry just seen this, in that case we should probably work out why it's not getting picked up...

.github/workflows/tests.yml
```diff
@@ -65,6 +65,8 @@ jobs:
           path: /usr/local/faasm/object
           key: ${{ runner.os }}-cache
       # --- Code build ---
+      - name: "Rerun cmake config"
```
Shillaker, line 68: Let's actually remove this, as you say, it already gets run :thinking:

csegarragonz: Already done it locally, forgot to push :man_facepalming:

cmake/ExternalProjects.cmake
```diff
@@ -20,8 +20,8 @@ FetchContent_Declare(wamr_ext
 )
 
 FetchContent_Declare(faabric_ext
-    GIT_REPOSITORY "https://github.com/faasm/faabric.git"
-    GIT_TAG "v0.0.12"
+    URL "https://github.com/faasm/faabric/archive/v0.0.13.tar.gz"
```
csegarragonz, line 23: I figured out that if our idea was to pin to a specific tag, we could as well just use the tag-specific `URL`, and avoid all the unexpected `git` behaviour like not fetching new tags/branches.

Additionally, by including the `URL_MD5` hash, we ensure the integrity of the content (not a real concern, but nice to have), as well as easily caching downloads. It is also [strongly recommended](https://cmake.org/cmake/help/latest/module/ExternalProject.html#module:ExternalProject) in the docs (look for `URL_HASH`).

To obtain such hash, I just use `md5sum` from the command line. I will add a line for it in the docs.

Shillaker: Hmm, so does this now pick up the changes if we change the tag? Have you tested it by changing to `0.0.12` and back? It seems weird that the URL approach would work properly while the Git one doesn't.

I'm very against the hash idea. The point of this approach is to make managing dependencies easier than using submodules (i.e. upgrades just require changing a CMake file), whereas with this approach we're adding another thing that has to be updated, and a set of extra commandline steps to generate that thing. This seems like more work than using submodules to me. 

Can you say which part of the docs' recommendation is relevant to us rather than just linking it? AFAICT there are two benefits, security and avoiding unnecessary downloads. The security angle isn't a concern for me, and I can't tell how important the unnecessary downloads bit really is, does it mean without the hash it'll constantly be downloading the .tar again?

Shillaker (changes requested):

Shillaker:
@csegarragonz have you tried using commit hashes instead of tags? I saw this and it does look like someone is saying their build carries on after it fails to fetch a tag: https://gitlab.kitware.com/cmake/cmake/-/issues/20677

csegarragonz:
After looking around for a bit, I think there's indeed a problem with the update command as presented in [this issue](https://gitlab.kitware.com/cmake/cmake/-/issues/20677) (`<C-f> The following is taken from`) with the update command not fetching tags. 

The logic contained in the [source code](https://gitlab.kitware.com/cmake/cmake/-/blob/v3.18.3/Modules/ExternalProject-gitupdate.cmake.in#L54) aims to minimise the network interactions when the source has already been fetched once. There are a couple of ways to circumvent it whilst still using the `GIT_` CMake submodule.

The first one is, as the problem lies in the default update routine, we can override it completely (forcing a re-download) by using `UPDATE_COMMAND ""` in the FetchContent_Declare. Alternatively, we can use `GIT_REMOTE_UPDATE_STRATEGY "CHECKOUT"`  which, according to the docs:
> GIT_REMOTE_UPDATE_STRATEGY <strategy>
When GIT_TAG refers to a remote branch, this option can be used to specify how the update step behaves. The <strategy> must be one of the following:
CHECKOUT
Ignore the local branch and always checkout the branch specified by GIT_TAG.

In essence, I don't think there's much difference between any of the two. Both seem to work, and  actually do very similar things. 

There fix for this issue [has already a PR](https://gitlab.kitware.com/cmake/cmake/-/merge_requests/5389) in Kitware's Git, but it's currently closed waiting for someone to write the test coverage :sweat: Worth keeping an eye on.

Shillaker (approved):

PR #361, opened 2020-11-24
Merge dev CLI and quick-start Docker Compose

Shillaker:
This merges the development docker-compose set-up and the quick-start docker-compose set-up (as the latter will rot if it's not the same as the former).

Changes:

- One docker-compose file for quick-start CLI and development CLI usage (note that the CLI container only depends on the redis containers, so if started alone will not create the worker and upload stuff)
- Fix broken CLI calls for uploading and invoking functions (these just didn't work before)
- Remove unnecessary parameters for passing around `host` and `port` in CLI functions
- Allow environment variables to override invoke/ upload hostnames (this is necessary to get the docker-compose configuration right)


docker-compose.yml
```diff
@@ -21,9 +21,9 @@ services:
     environment:
       - FUNCTION_STORAGE=local
       - LOG_LEVEL=info
-      - REDIS_STATE_HOST=redis-state
-      - REDIS_QUEUE_HOST=redis-queue
       - PYTHON_CODEGEN=off
+      - REDIS_QUEUE_HOST=redis-queue
+      - REDIS_STATE_HOST=redis-state
```
Shillaker, line 26: Alphabetical

faasmcli/faasmcli/tasks/call.py
```diff
@@ -17,8 +17,6 @@ def invoke(
     ctx,
     user,
     func,
-    host=None,
-    port=None,
```
Shillaker, line 21: Passing around the host and port is never used.

faasmcli/faasmcli/util/call.py
```diff
@@ -21,11 +23,55 @@ def _do_invoke(user, func, host, port, func_type, input=None):
     do_post(url, input, json=True)
 
 
+def _async_invoke(url, msg, headers=None, poll=False, host=None, port=None):
```
Shillaker, line 26: This is fixing a PEP warning where the complexity of the `invoke_impl` function was too high

faasmcli/faasmcli/util/endpoints.py
```diff
@@ -1,43 +1,33 @@
 from faasmcli.util.config import get_faasm_config
+from os import environ
 
 
-def get_upload_host_port(host_in, port_in):
+def _get_config_value(env_var, key, default_value):
+    """
+    Get the config value, allowing an environment variable to take precedence
+    """
```
Shillaker, line 8: Adding support for environment variables _and_ config file

bin/stop-cli.sh
```diff
@@ -1,15 +0,0 @@
...
-THIS_DIR=$(dirname $(readlink -f $0))
-PROJ_ROOT=${THIS_DIR}/..
-
-pushd ${PROJ_ROOT} > /dev/null
-
-# Stop all faasm-related containers
-echo "Stopping Faasm services"
-docker-compose -p faasm-dev -f docker/docker-compose-cli.yml stop
-
-popd > /dev/null
```
Shillaker, line 14: We can now just call `docker-compose down` or `docker ps -aq | xargs docker rm -f`

csegarragonz (approved):

PR #367, opened 2020-11-27
Python PEP8 checks

Shillaker:


bin/check_python.sh
```diff
@@ -0,0 +1,20 @@
+#!/bin/bash
+
+set -e
+
+THIS_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
```
csegarragonz, line 5: I am sorry to be so annoying with this, but could you add this `if` between these two lines:
```bash
if [ "$(ps -o comm= -p $$)" = "zsh" ]; then
    THIS_DIR="$( cd "$( dirname "${ZSH_ARGZERO}" )" >/dev/null 2>&1 && pwd )"
fi
```

csegarragonz (approved):
Other than the `if` thing for `zsh` looks good to me!

PR #371, opened 2020-12-01
Refactoring build type, and adding a utility compile

csegarragonz:
In this PR, I remove the need to pass the `FAASM_BUILD_FLAG=wasm` to different compilation targets. We already set the `CMAKE_SYSTEM_NAME` in the [toolchain file](https://github.com/faasm/faasm-toolchain/blob/master/WasiToolchain.cmake) so there's no need to carry the flag around.

Additionally, and to test that these changes did not break anything, I added a task to compile all functions needed for the tests, and updated the docs accordingly.


faasmcli/faasmcli/tasks/codegen.py
```diff
@@ -51,7 +51,3 @@ def local(ctx, wamr=False):
     _do_codegen_user("mpi", wamr=wamr)
     _do_codegen_user("omp", wamr=wamr)
     _do_codegen_user("python", wamr=wamr)
-
```
Shillaker, line 54: I can't see where these last two lines have moved to. We need to run the _wamr_ codegen for these two functions to run the WAMR tests

Shillaker (changes requested):

faasmcli/faasmcli/tasks/codegen.py
csegarragonz: oh man, thanks for spotting this, I completely overlooked the `wamr` flag :facepalm:

Shillaker (approved):

Shillaker:
@csegarragonz can we go further and completely remove the concept of `FAASM_BUILD_TYPE`? We can replace/ remove all occurrences shown here: https://github.com/faasm/faasm/search?q=FAASM_BUILD_TYPE

PR #376, opened 2020-12-08
Fix in MPI_Cart_create

csegarragonz:
In this PR I fix an issue with the current implementation of `MPI_Cart_create`.
This issue was creating an `OutOfBounds` exception when running `LAMMPS`, yet it
was not captured by the tests, hence I also add a regression test to fix it.

The problem was that I did not allocate memory for the newly created
communicator. This was making us copy an existing `faabric_communicator_t` into
an address pointed by an uninitialized pointer. This resulted in a
non-deterministic behaviour:
1. The tests pass as the pointer grabbed the value of the underlying memory
(`0`) and hence was building a communicator object at the base of the `wasm`
memory array.
2. On the other hand, in LAMMPS we tried to access a random value, which yielded
the `OutOfBounds` exception.

Note that in `OpenMPI` the communicator is allocated from the same
`ompi_cart_create` call (see [source code](https://github.com/open-mpi/ompi/blob/2d5e6668082c4099f4cd3bcb0c16d8866a38c306/ompi/mca/topo/base/topo_base_cart_create.c#L156))
so I opted to do the same.

Lastly, the test is somewhat naive, as it checks whether two consecutively
created communicators have the same address.

Closes faasm/experiment-lammps#2


csegarragonz:
Note to self: don't respect the 80 char limit when issuing PRs from GH cli.

src/wavm/mpi.cpp
```diff
@@ -1223,6 +1223,8 @@ WAVM_DEFINE_INTRINSIC_FUNCTION(env,
 /**
  * Makes a new communicator to which Cartesian topology information has been
  * attached.
+ * Note: In OpenMPI, memory is allocated from within the function call, that's
+ * why we allocate it here.
```
Shillaker, line 1227: It's not just because OpenMPI does it, it's because that's the MPI interface right?

Shillaker (approved):
LGTM apart from the small issue about the comment

PR #377, opened 2020-12-09
Faasm-side mplementation of MPI_Sendrecv

csegarragonz:
In this PR I include the `faasm`-side implementation of `MPI_Sendrecv`. The actual implementation was merged upstream to `faabric` in faasm/faabric#27 and patched in faasm/faabric#27. Check those PRs for the actual implementation details.

Additionally, I also include a feature test for a ring-like shift operation.

PS: this PR is blocked by faasm/faabric#37, once that's merged in we'll also have to update CMake's external dependency and make sure the toolchain is up to date.

PS2: This PR has ended up including some necessary changes to integrate the new faabric-as-a-library scheme. Most notably:
* Build all targets also in the `formatting` step (required by `clang-tidy`).
* Update `LD_LIBRARY_PATH` when running `codegen`.

Additionally, the new HTTP interface makes the test in `tests/test/upload/test_upload.cpp` fail to compile with the unresolved OpenSSL symbol error @Shillaker encountered before. I've commented the test out.


cmake/ExternalProjects.cmake
```diff
@@ -22,7 +22,7 @@ FetchContent_Declare(wamr_ext
 
 FetchContent_Declare(faabric_ext
     GIT_REPOSITORY "https://github.com/faasm/faabric.git"
-    GIT_TAG "7f8b43a8701fca33908cb3e7d18219c8d026689d"
+    GIT_TAG "origin/link-libraries"
```
csegarragonz, line 25: Will need to update this tag when faasm/faabric#47 is merged.

csegarragonz:
Tests fail with:
```
/build/faasm/bin/codegen_shared_obj: error while loading shared libraries: libpistache.so.0: cannot open shared object file: No such file or directory
```

I can reproduce this locally, but workaround it by running `sudo ldconfig`. Related with #382

.github/workflows/tests.yml
```diff
@@ -21,8 +21,8 @@ jobs:
         run: git fetch origin ${GITHUB_REF}:ci-branch
       - name: "Check out branch"
         run: git checkout --force ci-branch
-      - name: "Run CMake to regenerate compile commands"
-        run: inv -r faasmcli/faasmcli dev.cmake --build Release
+      - name: "Build the targets to update compile commands and dependencies"
```
csegarragonz, line 24: `clang-tidy` is requiring an increasing amount of compilation targets. In particular, in this PR I encountered:
* `clang-tidy` requiring protobuf's `*.pb.*` compiled files.
* `clang-tidy` requiring `pistache` headers installed with `faabric`.

As one could expect this to be a recurring issue, and given that we are already building the targets in a, concurrent, build, we decided offline to build the `dev.tools` target in the `formatting` workflow as well.

Shillaker (changes requested):
> Additionally, the new HTTP interface makes the test in tests/test/upload/test_upload.cpp fail to compile with the unresolved OpenSSL symbol error @Shillaker encountered before. I've commented the test out.

Are you sure this is definitely because of the new HTTP interface, or is it just because we've removed `ssl` and `crypto` as dependencies? It might be that other things depend on them, not just Faabric and Pistache (e.g. `cpprestsdk`). Remember that the upload server just uses `cpprestsdk`, and not anything HTTP-related from Faabric, so it's actually more likely that it's related to that.

Could you try adding back `ssl` and `crypto` as linked libraries in `storage` and/ or `upload`? I'd like to avoid commented out tests if possible. As a last resort, you can add `!mayfail` as you suggested although this shouldn't be necessary. I'm changing some HTTP-related stuff in #386 which may or may not be of interest.

src/upload/CMakeLists.txt
```diff
@@ -7,13 +7,13 @@ set(LIB_FILES
...
+find_package(cpprestsdk REQUIRED)
+
 target_link_libraries(upload_lib
         wasm
         wavmmodule
         faabric
-        ssl
-        crypto
-        cpprest
+        cpprestsdk::cpprest
```
Shillaker, line 16: Nice :ok_hand:

Shillaker (approved):

PR #379, opened 2020-12-10
Update clang-tidy to run on single file

Shillaker:
I had a problem with a clang-tidy file in the build that wasn't showing up in my local env, so I needed to run clang-tidy from the actual build container on a single file, and have it report the issues (rather than pipe the output to `/dev/null`). 

To do this, I modified the script to support accepting a single file, in which case it won't pipe the `clang-tidy` output to `/dev/null` (but it will continue to do so for directories or no argument).

Unfortunately I couldn't find a nice way to do the condition "if this then pipe", so I had to put the `clang-tidy` bit in a function that accepts an array input.

### DIFF
--- a/bin/run_clang_tidy.sh
+++ b/bin/run_clang_tidy.sh
@@ -4,27 +4,51 @@ set -e
 
 BUILD_PATH=/build/faasm/
 CONFIG=${FAASM_ROOT}/.clang-tidy
+DUMP_OUTPUT=true
 
-FILES=$(git ls-files {func,libs,src,tests}/{"*.h","*.cpp","*.c"})
+# Function to actually run clang-tidy
+function do_tidy {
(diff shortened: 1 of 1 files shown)


bin/run_clang_tidy.sh
```diff
@@ -4,27 +4,50 @@ set -e
...
 
-# Run clang-tidy on a set of files
-run-clang-tidy-10.py \
-    -config '' \
-    -j `nproc` \
-    -fix \
-    -format \
-    -style 'file' \
-    -p ${BUILD_PATH} \
-    -extra-arg-before='-I/usr/local/faasm/native/include' \
```
Shillaker, line 25: The inclusion of these headers is now done in the CMake config too

csegarragonz (approved):

PR #384, opened 2021-01-04
Fixes for Python ctypes, libffi, WASI capabilities, several extra tests

Shillaker:
This PR started out aiming to get Python's `ctypes` and libffi working, but getting there required a lot of bug fixing, debugging and testing to fix or rule out other issues. Summary below, sorry for the size of this:

_Features_

- Adding libffi calls to host interface
- Adding guard pages around memory for dynamically loaded modules
- Supporting `nullptr` passed to `dlopen` to get a reference to the main module
- Refactoring of dynamic loading logic (was hard to follow and potentially buggy before)
- Debug information dumps for wasm modules, filesystem and wasm environment variables

_Bugs and tests_

- Fixes and more rigorous tests for filesystem syscalls (esp. around dup and seek)
- Fixes and tests for WASI capabilities propagation
- Test for var-args handling
- Tests for python hashing, pickle, pyc handling and ctypes
- Deleted old `include/faasm` directory which has was moved to `faasm-toolchain` a while ago

All of this is also implicitly tested by just being able to run many of the Python functions successfully.


docker-compose.yml
```diff
@@ -78,6 +78,7 @@ services:
       - LD_LIBRARY_PATH=/build/faasm/third-party/lib:/usr/local/lib
     volumes:
       - .:/usr/local/code/faasm/
+      - /home/scs17/faasm-cpython:/usr/local/code/faasm-cpython
```
Shillaker, line 81: To be removed, dev hack

func/python/ctypes_check.py
```diff
@@ -0,0 +1,25 @@
+from ctypes import CDLL
```
Shillaker, line 1: Dev hack, to be removed

func/python/file_check.py
```diff
@@ -0,0 +1,12 @@
...
+
+def faasm_main():
+    with open(FILENAME, "rb") as fh:
+        contents = fh.read()
+
+    decoded = contents.decode(encoding="utf-8", errors="strict")
+
+    print(decoded)
+
+    return 0
```
Shillaker, line 12: Dev hack, can be removed

include/wasm/WasmModule.h
```diff
@@ -23,11 +23,13 @@
...
-// is 64kB so 480 pages is 30MB
-#define DYNAMIC_MODULE_STACK_SIZE 2 * ONE_MB_BYTES
-#define DYNAMIC_MODULE_HEAP_PAGES 480
+// Properties of dynamic modules
+// - NOTE - make sure these match any max-memory arguments passed when compiling
+// the modules.
+// Heap size must be wasm-module-page-aligned. One page is 64kB so 1920 pages
+// is 120MB
+#define DYNAMIC_MODULE_STACK_SIZE 4 * ONE_MB_BYTES
+#define DYNAMIC_MODULE_HEAP_PAGES 1920
```
Shillaker, line 32: May not be needed, just checking this wasn't an issue

include/wasm/WasmEnvironment.h
```diff
@@ -9,14 +10,22 @@ class WasmEnvironment
...
     uint32_t getEnvCount();
+
     uint32_t getEnvBufferSize();
 
-    const std::vector<std::string>& getVars();
+    std::vector<std::string> getVars();
+
+    std::string getEnv(const std::string& key);
+
+    void print();
```
Shillaker, line 26: Note to self: all this stuff may not be needed if we don't actually need to set `LC_CTYPES`

src/wavm/WAVMWasmModule.cpp
```diff
@@ -222,7 +221,7 @@ bool WAVMWasmModule::tearDown()
 
     dynamicPathToHandleMap.clear();
     for (auto const& m : dynamicModuleMap) {
-        dynamicModuleMap[m.first] = nullptr;
+        dynamicModuleMap[m.first].ptr = nullptr;
```
Shillaker, line 224: This is just to signal to WAVM that the module targeted by `ptr` can be GC-ed

include/wavm/WAVMWasmModule.h
```diff
@@ -138,21 +139,15 @@ class WAVMWasmModule final
     WAVM::Runtime::GCPointer<WAVM::Runtime::Instance> wasiModule;
     WAVM::Runtime::GCPointer<WAVM::Runtime::Instance> moduleInstance;
 
-    // Dynamic modules
-    int dynamicModuleCount = 0;
-    int nextMemoryBase = 0;
-    int nextStackPointer = 0;
-    int nextTableBase = 0;
```
Shillaker, line 145: Tracking the values of these `int` properties was difficult, so I've refactored the logic into the `LoadedDynamicModule` class

include/storage/FileSystem.h
```diff
@@ -32,10 +32,15 @@ class FileSystem
...
     static void clearSharedFiles();
 
+    std::string getPathForFd(int fd);
+
   private:
     int nextFd;
+
     std::unordered_map<int, storage::FileDescriptor> fileDescriptors;
 
+    std::unordered_map<int, std::string> fdNames;
```
Shillaker, line 42: This is to make the logging around filesystem calls easier. Before it was just logging integers for file descriptors which was a little difficult to follow. Now it logs the path of the file by looking it up in this map.

faasmcli/faasmcli/tasks/run.py
```diff
@@ -23,7 +21,6 @@ def run(ctx, user, function, repeats=1, wamr=False):
         args.append(str(repeats))
 
     wasm_vm = "wamr" if wamr else "wavm"
-    env = copy(environ)
-    env.update({"WASM_VM": wasm_vm})
+    extra_env = {"WASM_VM": wasm_vm}
 
-    run_command("simple_runner", args)
+    run_command("simple_runner", args, extra_env=extra_env)
```
Shillaker, line 26: This wasn't actually being passed through before.

func/demo/backtrace.cpp
```diff
@@ -0,0 +1,16 @@
...
+{
+    __faasm_backtrace(0);
+}
+
+int main(int argc, char* argv[])
+{
+    someFunction();
+
+    return 0;
+}
```
Shillaker, line 16: This just checks the backtrace function doesn't fail.

func/demo/filedescriptor.cpp
```diff
@@ -0,0 +1,198 @@
...
+
+#include <cerrno>
+#include <cstdio>
+#include <cstdlib>
+#include <cstring>
+#include <fcntl.h>
+#include <string>
+
+#include <unistd.h>
+
```
Shillaker, line 11: I should have added a function like this when I wrote all the filesystem syscalls, but I did it in a bit of a rush (and it's regularly come back to bite me). It's meant to test the low-level handling of file descriptors, mainly `dup`, `seek` and `write`

func/demo/va_arg.cpp
```diff
@@ -0,0 +1,83 @@
+#include <cstdarg>
+#include <cstdio>
+#include <cstdlib>
+#include <cstring>
+#include <faasm/compare.h>
+#include <vector>
+
```
Shillaker, line 7: Var-args isn't something we have modified or implemented, but is worth testing just to be sure it works (as it looked like this was the problem for a while).

func/python/env_check.py
```diff
@@ -0,0 +1,25 @@
+import os
+
+
```
Shillaker, line 3: This test was missing before, just checking the Python interaction with the environment variables (again, something I thought was a problem working through the bug)

func/python/hash_check.py
```diff
@@ -0,0 +1,14 @@
+import hashlib
```
Shillaker, line 1: Hashlib is another internal CPython library that lives in `lib-dynload`, hence I needed to check whether it was _just_ `ctypes`, or all `lib-dynload` libraries.

func/python/hello.py
```diff
@@ -1,11 +1,11 @@
 import sys
 
-from pyfaasm.core import set_output
+from pyfaasm.core import write_output
```
Shillaker, line 3: The `pyfaasm` interface has changed slightly as part of the move to `ctypes`.

func/python/pickle_check.py
```diff
@@ -0,0 +1,19 @@
+import pickle
```
Shillaker, line 1: `pickle` is another `lib-dynload` library (i.e. uses a shared object under the hood), so it's important to test.

func/python/pyc_check.py
```diff
@@ -0,0 +1,44 @@
+import bisect
+import dis
+import marshal
+import io
+
```
Shillaker, line 5: For a long time I thought pyc files were the problem, so this checks they are read and parsed properly.

func/python/pyinit.c
```diff
@@ -126,8 +126,9 @@ void setUpPyNumpy()
     fprintf(devNull, "%Le\n", (long double)strtoull("1e100", NULL, 0));
 
     // Strings
-    FORCE_LINK(strpbrk);
+    FORCE_LINK(strcat);
     FORCE_LINK(strcspn);
+    FORCE_LINK(strpbrk);
```
Shillaker, line 131: This whole file is there to ensure these functions are compiled into the static binary, so this is just adding a couple that were missing.

include/storage/FileDescriptor.h
```diff
@@ -109,7 +111,7 @@ class FileDescriptor
 
     bool mkdir(const std::string& dirPath);
 
-    uint16_t seek(uint64_t offset, int wasiWhence, uint64_t* newOffset);
+    uint16_t seek(int64_t offset, int wasiWhence, uint64_t* newOffset);
```
Shillaker, line 114: These offsets can be negative, hence an unsigned int is not a good choice...

include/wavm/LoadedDynamicModule.h
```diff
@@ -0,0 +1,39 @@
+#pragma once
+
```
Shillaker, line 2: This class is used to encapsulate the properties of loaded dynamic modules. Once the module is created we don't really need to care about this, but it's really useful for debugging and sanity checking.

include/wavm/WAVMWasmModule.h
```diff
@@ -194,6 +192,8 @@ class WAVMWasmModule final
     void prepareOpenMPContext(const faabric::Message& msg);
 
     std::unique_ptr<openmp::PlatformThreadPool> OMPPool;
+
+    uint32_t createMemoryGuardRegion();
```
Shillaker, line 196: For a while I thought ther was some sort of memory corruption issue, so adding guard regions around the dynamic modules makes sense. This way we should be able to catch stack-overflows in those modules.

src/storage/FileDescriptor.cpp
```diff
@@ -689,4 +688,26 @@ void FileDescriptor::setActualRights(uint64_t rights, uint64_t inheriting)
...
+    path = other.path;
+    rightsSet = other.rightsSet;
+    actualRightsBase = other.actualRightsBase;
+    actualRightsInheriting = other.actualRightsInheriting;
+
+    dirContentsLoaded = other.dirContentsLoaded;
+    dirContents = other.dirContents;
+    dirContentsIdx = other.dirContentsIdx;
+
+    return linuxFd;
```
Shillaker, line 710: The implementation of `dup` was a bit of a hack before, and I _think_ it worked ok, but this is more explicit.

src/wavm/WAVMWasmModule.cpp
```diff
@@ -188,11 +184,14 @@ void WAVMWasmModule::clone(const WAVMWasmModule& other)
 
         // Remap dynamic modules
         // TODO - double check this works
+        lastLoadedDynamicModuleHandle = other.lastLoadedDynamicModuleHandle;
```
Shillaker, line 187: This `lastLoadedDynamicModuleHandle` is unfortunately needed for linking (we load the module, then some time later WAVM tries to link it, at which point we need to revisit the thing we just created. This happens in the `resolve` method).

src/wavm/WAVMWasmModule.cpp
```diff
@@ -643,50 +682,99 @@ int WAVMWasmModule::dynamicLoadModule(const std::string& path,
...
+    int thisHandle;
+    if (path.empty()) {
+        logger->warn("Dynamic linking main module");
+        return MAIN_MODULE_DYNLINK_HANDLE;
+    } else if (boost::filesystem::is_directory(path)) {
+        logger->warn("Dynamic linking a directory {}", path);
+        return 0;
+    } else if (!boost::filesystem::exists(path)) {
+        logger->warn("Dynamic module {} does not exist", path);
+        return 0;
```
Shillaker, line 707: The previous implementation didn't conform to the `dlopen` interface, where you are allowed to pass `nullptr` to get a reference to the currently executing shared object. In our case, this currently executing shared object is just the function itself, so we use this `MAIN_MODULE_DYNLINK_HANDLE` to refer back to the main module.

src/wavm/WAVMWasmModule.cpp
```diff
@@ -905,7 +993,7 @@ U32 WAVMWasmModule::mmapFile(U32 fd, U32 length)
 
 U32 WAVMWasmModule::allocateThreadStack()
 {
-    return this->mmapMemory(THREAD_STACK_SIZE);
+    return mmapMemory(THREAD_STACK_SIZE);
```
Shillaker, line 996: We should only use `this->` where it's necessary for clarity.

src/wavm/WAVMWasmModule.cpp
```diff
@@ -1494,4 +1592,89 @@ std::unique_ptr<openmp::PlatformThreadPool>& WAVMWasmModule::getOMPPool()
 {
     return OMPPool;
 }
+
+void WAVMWasmModule::printDebugInfo()
+{
```
Shillaker, line 1597: I've added `printDebugInfo` to a few classes now. It's useful for spitting out a dump of information when things fail, as it can be hard to work out what's going on with just the wasm stacktrace (or only a Python stack trace!)

src/wavm/WAVMWasmModule.cpp
```diff
@@ -1494,4 +1592,89 @@ std::unique_ptr<openmp::PlatformThreadPool>& WAVMWasmModule::getOMPPool()
...
+    if (res != 0) {
+        logger->error("Failed to create memory guard: {}",
+                      std::strerror(errno));
+        throw std::runtime_error("Failed to create memory guard");
+    }
+
+    logger->debug(
+      "Created guard region {}-{}", wasmOffset, wasmOffset + regionSize);
+
+    return wasmOffset;
```
Shillaker, line 1678: Difficult to test that these guard regions actually work, as trying to write to them causes a `SIGSEGV`

src/wavm/dynlink.cpp
```diff
@@ -54,10 +70,16 @@ WAVM_DEFINE_INTRINSIC_FUNCTION(env,
...
 
-    // Ignore
-    return 0;
+    wasm::WAVMWasmModule* module = getExecutingWAVMModule();
+    uint32_t wasmStrPtr = module->mmapMemory(errorMessage.size());
+
+    char* strPtr = &Runtime::memoryRef<char>(module->defaultMemory, wasmStrPtr);
+    ::strcpy(strPtr, errorMessage.c_str());
+
+    return wasmStrPtr;
```
Shillaker, line 82: `dlerror` wasn't actually implemented before, so errors with dynamic linking where ignored (leading to some confusion...)

src/wavm/dynlink.cpp
```diff
@@ -67,4 +89,170 @@ WAVM_DEFINE_INTRINSIC_FUNCTION(env, "dlclose", I32, dlclose, I32 handle)
     // Ignore
     return 0;
 }
+
+// ----------------------------------
+// libffi - see https://github.com/faasm/libffi/
```
Shillaker, line 94: Adding libffi was the original aim of this PR, as it's needed to support ctypes.

src/wavm/io.cpp
```diff
@@ -338,23 +339,39 @@ WAVM_DEFINE_INTRINSIC_FUNCTION(wasi,
...
-    return __WASI_ESUCCESS;
+    Runtime::memoryRef<uint64_t>(getExecutingWAVMModule()->defaultMemory,
+                                 resBytesWrittenPtr) = bytesWritten;
+
+    delete[] nativeIovecs;
+
+    if (bytesWritten < 0) {
+        return storage::errnoToWasi(errno);
+    } else {
+        return __WASI_ESUCCESS;
```
Shillaker, line 373: This function was ignoring errors, and not correctly setting the number of bytes written.

func/demo/getenv.cpp
```diff
@@ -21,7 +21,7 @@ int main(int argc, char* argv[])
 {
     // Check predefined var
     bool failed = false;
-    failed |= checkPredefined("LC_CTYPE", "en_GB.UTF-8");
+    failed |= checkPredefined("PYTHONNOUSERSITE", "on");
```
Shillaker, line 24: Small tweaks to the Python env.

func/python/dict_state.py
```diff
@@ -1,9 +1,9 @@
 import pickle
 from pyfaasm.core import (
-    get_state,
-    get_state_size,
-    set_state,
-    chain_this_with_input,
+    read_state,
+    read_state_size,
+    write_state,
+    chain,
```
Shillaker, line 6: Quite a few Python functions have been updated to fit with the new pyfaasm interface

include/ir_cache/IRModuleCache.h
```diff
@@ -36,6 +36,10 @@ class IRModuleCache
                                  const std::string& func,
                                  const std::string& path);
 
+    size_t getSharedModuleDataSize(const std::string& user,
+                                   const std::string& func,
+                                   const std::string& path);
```
Shillaker, line 41: This is here to give a bit more detail on the memory layout of shared modules (which is now a lot more explicit in the dynamic module loading)

src/storage/FileSystem.cpp
```diff
@@ -56,6 +58,15 @@ int FileSystem::getNewFd()
...
 
+std::string FileSystem::getPathForFd(int fd)
+{
+    if (fileDescriptors.count(fd) == 0) {
+        return "";
+    }
+
+    return fileDescriptors[fd].getPath();
+}
+
```
Shillaker, line 69: This function helps with giving more meaningful logging during filesystem calls (rather than just printing out the integer handle)

src/wavm/io.cpp
```diff
@@ -849,6 +863,19 @@ WAVM_DEFINE_INTRINSIC_FUNCTION(env,
 // Unsupported
 // -----------------------------
 
+WAVM_DEFINE_INTRINSIC_FUNCTION(env,
+                               "__small_sprintf",
```
Shillaker, line 867: This is an Emscripten function but doesn't actually get called so it's not worth implementing

include/faasm/faasm.imports
```diff
@@ -1,33 +0,0 @@
-__faasm_push_state
```
Shillaker, line 1: This moved to `faasm-toolchain` (along with everything else in the `include/faasm` directory), but it escaped deletion from this repo.

It only became a problem when adding a new function to `host_interface.h`, and the linting was picking up the version in this repo, rather than the one installed by `faasm-toolchain`

func/demo/filedescriptor.cpp
```diff
@@ -0,0 +1,265 @@
...
+int main(int argc, char* argv[])
+{
+    // Start off with the basic permissions
+    validatePermissions();
+
+    const char* filePath = "fileA.txt";
+    deleteFile(filePath);
+
+    std::string contentA = "This is my file.\n\
```
csegarragonz, line 126: https://youtu.be/7Qn-jH0gl14?t=33

src/wavm/WAVMWasmModule.cpp
csegarragonz: I understand this actually works now right? Could we remove the `TODO`?

csegarragonz (approved):

src/wavm/WAVMWasmModule.cpp
Shillaker: Good point, yes I'm pretty sure this is covered by a test somewhere

PR #389, opened 2021-01-27
Horovod example

Shillaker:


Shillaker:
This is no longer needed.

PR #392, opened 2021-03-08
Update kubernetes and knative logic

csegarragonz:
This PR now contains the necessary changes to move from `knative` `0.13.0` to `0.21.0`. This was needed as the original installation scripts weren't working.

I also include an extra file in the `faasm.ini` to contain the `host` file which may change depending on the underlying hardware.

Afer merging, updating the `faasm.ini` and running:
```
inv upload demo hello
inv invoke demo hello
```
should run as expected.

### DIFF
--- a/docs/kubernetes.md
+++ b/docs/kubernetes.md
@@ -30,21 +30,10 @@ To set up Faasm on [GKE](https://console.cloud.google.com/kubernetes) you can do
 To set up Faasm on AKS, you can do the following:
 - Set up an account and install the [`az`](https://docs.microsoft.com/en-us/cli/azure/install-azure-cli) client.
 - Create a resource group with sufficient quota.
-- Run:
-```bash
-az aks create \
-    --resource-group <YOUR_RESOURCE_GROUP> \
-    --name <CLUSTER_NAME> \
-    --node-count <CLUSTER_NODE_COUNT> \
-    --node-vm-size <VM_SIZE> \
-    --generate-ssh-keys
-```
-- Lastly, to update your local config run:
-```bash
-az aks get-credentials \
-    --resource-group <YOUR_RESOURCE_GROUP> \
-    --name <CLUSTER_NAME>
-```
+- Create an AKS cluster under said resource group.
+- Aim for >=4 nodes with more than one vCPU
+- Set up the local `kubectl` via the `get credentials` command
+- Install Knative as described below
 
 ### Bare metal
 
@@ -72,7 +61,7 @@ If you are running a local cluster or one on AKS, you will have to update the `e
 field in the `upload-service.yml`.
 You can query it with:
 ```bash
-kubectl get svc istio-ingressgateway -n istio-system -o=jsonpath='{.status.loadBalancer.ingress[0].ip}'
+./bin/knative_route.sh
 ```
 
 ### Faasm
--- a/docs/kubernetes.md
+++ b/docs/kubernetes.md
@@ -59,10 +59,7 @@ kubectl get pods -n knative-serving
 
 If you are running a local cluster or one on AKS, you will have to update the `externalIPs`
 field in the `upload-service.yml`.
-You can query it with:
-```bash
-./bin/knative_route.sh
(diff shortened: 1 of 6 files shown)


docs/kubernetes.md
```diff
@@ -25,6 +25,27 @@ To set up Faasm on [GKE](https://console.cloud.google.com/kubernetes) you can do
...
+    --node-count <CLUSTER_NODE_COUNT> \
+    --node-vm-size <VM_SIZE> \
+    --generate-ssh-keys
+```
+- Lastly, to update your local config run:
+```bash
+az aks get-credentials \
+    --resource-group <YOUR_RESOURCE_GROUP> \
+    --name <CLUSTER_NAME>
+```
```
Shillaker, line 47: Let's avoid specifying how people create a cluster (as with the GKE instructions above), you can like to the AKS docs if you think they're useful, but really all we care about is that someone has a cluster where they can run `kubectl`. 

You can add any recommended resource levels like in the GKE instructions above (e.g. use at least 4 hosts and nothing smaller than XYZ VMs)

docs/kubernetes.md
```diff
@@ -45,7 +66,14 @@ inv knative.install
...
 kubectl get pods -n knative-serving
-```  
+```
+
+If you are running a local cluster or one on AKS, you will have to update the `externalIPs`
+field in the `upload-service.yml`.
+You can query it with:
+```bash
+kubectl get svc istio-ingressgateway -n istio-system -o=jsonpath='{.status.loadBalancer.ingress[0].ip}'
+```
```
Shillaker, line 76: It seems like we have two places for this configuration now, the `knative_route.sh` script and the README. Could you add this line to the `knative_route.sh` script and make it print out there?

If it's not in a script, a command like this tends to rot over time.

Shillaker (changes requested):

docs/kubernetes.md
Shillaker: If you want to keep a specific script for the faasm experiments, you could create one in the `experiments-base` repo, e.g. `bin/provision_aks_cluster.sh`

faasmcli/faasmcli/util/env.py
```diff
@@ -30,7 +30,7 @@ def _get_dir(variable, default):
 
 FAASM_BUILD_DIR = _get_dir("FAASM_BUILD_DIR", "/build/faasm")
 
-FAASM_CONFIG_FILE = join(FAASM_HOME, "faasm.ini")
+FAASM_CONFIG_FILE = join(PROJ_ROOT, "faasm.ini")
```
csegarragonz, line 33: I think this file is generally outdated, but this is the only required change now.

faasmcli/faasmcli/tasks/knative.py
```diff
@@ -238,15 +238,40 @@ def _deploy_knative_fn(
...
+def uninstall(ctx):
+    """
+    Uninstall knative from an existing k8s cluster
+    """
+    specs = [
+        ["knative/net-istio", "net-istio.yaml"],
+        ["knative/net-istio", "istio.yaml"],
+        ["knative/serving", "serving-core.yaml"],
+        ["knative/serving", "serving-crds.yaml"],
+    ]
```
Shillaker, line 270: This list of 4 values is duplicated, could we perhaps have a constant called `KNATIVE_SPECS` and just add `serving-default-domain.yaml` in the `install` task? Why don't we call delete on `serving-default-domain.yaml` as well?

Shillaker (commented):
One small formatting change in the Python scripts but otherwise LGTM

faasmcli/faasmcli/tasks/knative.py
csegarragonz: The reason was that the lists are not traversed in the same order, as otherwise I would sometimes get errors complaining that certain component were not there.

And the `serving-default-domain.yml` is a one-off task, and creates no resources.

Shillaker (approved):

PR #393, opened 2021-03-10
Bump Faabric + Changes in CMake FetchContent

csegarragonz:
This PR bumps the `faabric` version to point to the latest ref. The logging enhancements we implemented in faasm/faabric#56 broke things downstream. This PR fixes them, and some other things that sprung algon the way.

I removed the sgx tests when sgx was not found.

I also updated the toolchain to have the latest libraries installed, and I've bumped faabric version accordingly (faasm/faasm-toolchain#29).

Additionally, this has helped me catch an unexpected behaviour of CMake's `FetchContent_Declare`. In particular, the options we pass via `CMAKE_ARGS -D__FLAGS__` are _not_ passed to the dependency's configure step. For references see this [question](https://cmake.org/pipermail/cmake/2019-March/069206.html) and [this other one](https://stackoverflow.com/questions/62101576/using-fetchcontent-declare-together-with-cmake-args).

In fact, we have been setting the `faabric_tests` build to off, but in reality we link against `faabric_test_utils` in faasm's tests, what means that the tests were actually being built.


src/wavm/CMakeLists.txt
```diff
@@ -34,5 +34,6 @@ set(LIB_FILES
         )
 
 faasm_private_lib(wavmmodule "${LIB_FILES}")
+target_compile_options(wavmmodule PRIVATE "-D__wasm__")
```
csegarragonz, line 37: I am not sure why the `define` in `wasi-libc` does not get here, but it does not. I am also concerned that other compile options we pass in `cmake/ExternalProjects.cmake` like `WAVM_ENABLE_UNWIND=ON` are not passed to the compiler when _actually_ building `WAVM`.

Shillaker: We're not building this to wasm, so why do you expect `-D__wasm__` to be defined?

CMakeLists.txt
```diff
@@ -13,6 +13,7 @@ option(WAMR_INTERPRETER_MODE "Turns interpreter on / off (vs. AoT)" OFF)
 
 # SGX functionality
 # These options customise the SGX features _provided_ SGX is found
+option(FAASM_SGX_ENABLED "Enable SGX mode" OFF)
```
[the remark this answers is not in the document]
Shillaker: I thought that `find_pacakage(SGX)` would just set `SGX_FOUND=false` if SGX wasn't installed, and that's how you do an SGX-free build. This autodetects the environment rather than having to change the option flag, which seems a bit nicer.

tests/test/CMakeLists.txt
```diff
@@ -1,4 +1,10 @@
 file(GLOB_RECURSE TEST_FILES ${CMAKE_CURRENT_LIST_DIR} test_*.cpp)
+file(GLOB_RECURSE SGX_TEST_FILES sgx test_*.cpp)
+
+# Remove SGX tests if SGX disabled or not found
+if (NOT (FAASM_SGX_ENABLED AND SGX_FOUND))
+    list(REMOVE_ITEM ${TEST_FILES} ${SGX_TEST_FILES})
+endif()
```
Shillaker, line 7: As per comment above, can't we just rely on `SGX_FOUND`? Why do we need the extra `FAASM_SGX_ENABLED` variable?

Shillaker (changes requested):

tests/test/CMakeLists.txt
Shillaker: Ah ok, I would say the normal behaviour is for builds to look for what they need on a system, then build with it if they find it (e.g. something will build with CUDA or BLAS support if it finds that library). Was it failing with your SGX install? I'd rather we keep compiling SGX support by default otherwise the code will rot. 

If you do still want to turn it off, I think the option should be to opt out (rather than opt in), i.e. default to `FAASM_SGX_ENABLED=ON`

src/wavm/CMakeLists.txt
csegarragonz: I keep getting confused with this. Thanks.

Wrt `WAVM_ENABLE_UNWIND` it is set to ON by default [here](https://github.com/faasm/WAVM/blob/95fd0c4374f68570cfd2a7e51b227bf593d5eec7/CMakeLists.txt#L129), so I think there's no need to pass the extra argument.

tests/test/CMakeLists.txt
csegarragonz: Ok, I see what you mean. SGX build is also sorted now, so I will revert to the previous behaviour (with the caveat of the removal of `sgx` tests if `NOT SGX_FOUND`.

src/wavm/CMakeLists.txt
Shillaker: Ok great, so we can remove this definition right?

Shillaker (approved):

PR #394, opened 2021-03-12
Faabric submodule and repo renaming

Shillaker:


CMakeLists.txt
```diff
@@ -143,6 +143,10 @@ add_definitions(${LLVM_DEFINITIONS})
 # Third party deps
 include(cmake/ExternalProjects.cmake)
 
+# Faabric
+add_subdirectory(faabric)
+include_directories(${CMAKE_CURRENT_LIST_DIR}/faabric/include)
```
Shillaker, line 148: This is exactly what we were just talking about with Jakub :see_no_evil:. We should add this as a public target include to the faabric target, but I'll fix _all_ these CMake issues in one go.

kubasz:
It might be worth it adding the submodule presence check/automatic update CMake code as described here: <https://cliutils.gitlab.io/modern-cmake/chapters/projects/submodule.html>

The `if not exists` check with `fatal_error` would have to be replicated once per submodule, and the init/update `execute_process` should happen only once.

Shillaker:
Awesome, thanks @kubasz , I'm building a list of low-hanging fruit we can fix with the CMake set-up, then I'll put it all into a single PR once I've finished rearranging the dev environment.

csegarragonz (approved):

PR #395, opened 2021-03-12
Moving function code out of this repo

Shillaker:
This PR is mostly deleting code that now lives in other repos. 

However, there are still a couple of changes worth checking:

- Github Actions - added steps to use `cpp` and `python` containers to create wasm and runtime files. These are then uploaded as artifacts, and downloaded by the container running the tests. This is pretty gross and I would like to have done it through mounted volumes, but each job executes in a different VM and the only way to share directories is via artifacts. 
- Merging in everything from the `development` repo (esp. [docker-compose file](https://github.com/faasm/faasm/blob/func-move/docker-compose-dev.yml) and [development docs](https://github.com/faasm/faasm/blob/func-move/docker-compose-dev.yml))


faasmcli/faasmcli/tasks/upload.py
```diff
@@ -1,36 +1,22 @@
-import multiprocessing
-from functools import partial
-from os import makedirs, listdir
+from os import makedirs
```
Shillaker, line 1: Faasm no longer has the concept of functions held in this repo, so the upload task is now much simpler (just points at a file)

bin/create_venv.sh
```diff
@@ -0,0 +1,31 @@
+#!/bin/bash
```
Shillaker, line 1: From development repo

bin/cli.sh
```diff
@@ -1,38 +1,75 @@
 #!/bin/bash
-
 set -e
```
Shillaker, line 2: From `development` repo

bin/refresh_local.sh
```diff
@@ -0,0 +1,34 @@
+#!/bin/bash
+
```
Shillaker, line 2: From `development` repo

docker-compose-dev.yml
```diff
@@ -0,0 +1,82 @@
+version: "3"
```
Shillaker, line 1: This is the docker-compose file that pulls together all the projects, basically just copied from the `development` repo

docs/development.md
```diff
@@ -1,119 +1,125 @@
 # Faasm Development
```
Shillaker, line 1: Worth looking at the final version rather than the diff: https://github.com/faasm/faasm/blob/func-move/docs/development.md

faasmcli/faasmcli/tasks/codegen.py
```diff
@@ -3,8 +3,15 @@
...
+from os.path import join
 
-from faasmcli.util.codegen import find_codegen_func
+from faasmcli.util.codegen import find_codegen_func, find_codegen_shared_lib
+from faasmcli.util.env import FAASM_RUNTIME_ROOT
+
+LIB_FAKE_FILES = [
+    join(FAASM_RUNTIME_ROOT, "lib", "libfakeLibA.so"),
+    join(FAASM_RUNTIME_ROOT, "lib", "libfakeLibB.so"),
+]
```
Shillaker, line 14: Extra step for doing codegen on test library

faasmcli/faasmcli/tasks/python.py
```diff
@@ -37,27 +37,6 @@ def clear_runtime_pyc(ctx):
...
-
-    # Create a tmp directory
-    run(
-        "mkdir -p {}".format(join(PY_RUNTIME_ROOT, "tmp")),
-        shell=True,
-        check=True,
-    )
-
-    # Run codegen
-    codegen(ctx)
```
Shillaker, line 58: Was no longer used anyway

.github/workflows/tests.yml
```diff
@@ -35,8 +31,67 @@ jobs:
       - name: "Run C/C++ linter"
         run: ./bin/run_clang_tidy.sh
 
+  cpp-funcs:
+    if: github.event.pull_request.draft == false
+    runs-on: ubuntu-20.04
+    container:
+      image: faasm/cpp-sysroot:0.0.22
```
Shillaker, line 38: Annoyingly we have another two versions to maintain in here now, but we can worry about that later

.github/workflows/tests.yml
```diff
@@ -35,8 +31,67 @@ jobs:
...
+    defaults:
+      run:
+        working-directory: /__w/faasm/faasm/clients/cpp
```
Shillaker, line 41: Github actions `checkout` action is much quicker than our branch approach and a bit more readable, but it checks it out to `/__w/<org>/<repo>/`

.gitignore
```diff
@@ -1,4 +1,5 @@
 # Faasm
+dev/
```
Shillaker, line 2: In the new docker-compose set-up we put all the development-related stuff in `dev`

docker/base.dockerfile
```diff
@@ -1,5 +1,4 @@
 # Stages to extract toolchain and sysroot
-FROM faasm/cpp-sysroot:0.0.21 as sysroot
 FROM faasm/cpython:0.0.10 as python
```
Shillaker, line 2: Unfortunately we still depend on the Python files at runtime, so have to copy them in using a multi-stage build

docs/development.md
csegarragonz: Looking :new: !

csegarragonz (approved):

PR #397, opened 2021-03-22
Bulk-scheduling update

Shillaker:
The main change in here is to have the pthread and OpenMP implementations use the batch scheduling now supported in Faabric: https://github.com/faasm/faabric/pull/61 , but I encountered a few errors along the way related to snapshot and restore, so I've had to fix them too.

In the past, we had two thread modes, one where every thread would be executed as a separate function, and one where every thread would be executed locally on a single machine. Faabric's batch scheduling now means we can do a mix, where as many threads will be executed locally as possible, then the rest will be distributed.


docs/openmp.md
```diff
@@ -1,22 +1,11 @@
...
-
-## OpenMP functions
-
-To compile, upload and run the pre-built OpenMP demo function:
-
-```
-inv compile omp hellomp
-inv upload omp hellomp
-inv invoke omp hellomp
-```
```
Shillaker, line 21: Now moved to cpp repo

include/module_cache/WasmModuleCache.h
```diff
@@ -16,8 +16,6 @@ class WasmModuleCache
     std::shared_mutex mx;
     std::unordered_map<std::string, wasm::WAVMWasmModule> cachedModuleMap;
 
-    std::string getCachedModuleKey(const faabric::Message& msg);
-
```
Shillaker, line 20: Inlined this function so not needed.

include/wasm/WasmModule.h
```diff
@@ -129,6 +133,9 @@ class WasmModule
     int stdoutMemFd;
     ssize_t stdoutSize;
 
+    std::mutex moduleMemoryMutex;
+    std::mutex moduleStateMutex;
```
Shillaker, line 137: These two mutexes are required as things can access module memory directly, but state stuff may also need to access memory too (while already holding the state lock). I could have used a single module-level recursive lock, but I think this is clearer and easier to debug when things go wrong.

src/module_cache/WasmModuleCache.cpp
```diff
@@ -70,33 +60,39 @@ wasm::WAVMWasmModule& WasmModuleCache::getCachedModule(
...
+                // Clone the special module from the base one
+                specialModule = baseModule;
+
+                // Restore the special module
+                specialModule.restoreFromState(snapshotKey, msg.snapshotsize());
+
+                // Write memory to fd to support copy-on-write cloning
+                int fd = memfd_create(snapshotKey.c_str(), 0);
+                specialModule.writeMemoryToFd(fd);
+            }
```
Shillaker, line 91: This is basically just a rearrange to make it more readable and easier to step through

src/wasm/chaining_util.cpp
```diff
@@ -95,14 +88,6 @@ int spawnChainedThread(const std::string& snapshotKey,
...
     // Schedule the call
     sch.callFunction(call);
-    faabric::util::getLogger()->debug(
-      "Chained thread {} ({}) -> {} {}({}) ({})",
-      origStr,
-      faabric::util::getSystemConfig().endpointHost,
-      chainedStr,
-      funcPtr,
-      argsPtr,
-      call.scheduledhost());
```
Shillaker, line 105: The scheduler provides enough logging for this now.

src/wavm/openmp.cpp
```diff
@@ -363,13 +365,13 @@ WAVM_DEFINE_INTRINSIC_FUNCTION(env,
     OMP_FUNC_ARGS(
       "__kmpc_fork_call {} {} {} {}", locPtr, argc, microtaskPtr, argsPtr);
 
+    auto conf = faabric::util::getSystemConfig();
```
Shillaker, line 368: See PR description, this function has pretty much been rewritten.

src/wavm/threads.cpp
```diff
@@ -86,16 +97,58 @@ WAVM_DEFINE_INTRINSIC_FUNCTION(env,
                   entryFunc,
                   argsPtr);
 
+    faabric::Message* originalCall = getExecutingCall();
```
Shillaker, line 100: Quite a big rewrite of this function too (`pthread_create`).

tests/test/faaslet/test_pool.cpp
```diff
@@ -34,11 +34,9 @@ static void tearDown()
     cleanSystem();
 }
 
-TEST_CASE("Test worker initially pre-warmed", "[faaslet]")
+TEST_CASE("Test faaslet not initially bound", "[faaslet]")
 {
     setUp();
-
-    FaasletPool pool(1);
```
Shillaker, line 41: In the past these tests had to create an instance of `FaasletPool`, but don't need to any more so I've removed them

tests/test/faaslet/test_pool.cpp
```diff
@@ -230,56 +224,4 @@ TEST_CASE("Test pool accounting", "[faaslet]")
     w1.bindToFunction(call);
     REQUIRE(pool.getThreadCount() == 2);
 }
-
-TEST_CASE("Test worker lifecycle interacts with scheduler", "[faaslet]")
```
Shillaker, line 234: This is basic functionality associated with Faabric executors, and tested in Faabric itself.

tests/test/main.cpp
```diff
@@ -4,6 +4,21 @@
...
+    using TestEventListenerBase::TestEventListenerBase;
+
+    void testCaseStarting(Catch::TestCaseInfo const& testInfo) override
+    {
+        auto logger = faabric::util::getLogger();
+        logger->debug("---------------------------------------------");
+        logger->debug("TEST: {}", testInfo.name);
+        logger->debug("---------------------------------------------");
+    }
+};
```
Shillaker, line 18: As in Faabric, I've added this to print test names. Without it its really difficult to parse the stream of test logging.

src/wavm/openmp.cpp
```diff
@@ -378,188 +380,154 @@ WAVM_DEFINE_INTRINSIC_FUNCTION(env,
...
+
+        // Snapshot details
+        call.set_snapshotkey(snapshotKey);
+        call.set_snapshotsize(snapshotSize);
+
+        // Function pointer
+        call.set_funcptr(microtaskPtr);
+
+        // Args
+        // TODO - is the order of these right? Does protobuf push to the back or
```
csegarragonz, line 421: Is this solved? Seems like a breaking choice.

csegarragonz (approved):

csegarragonz:
Just a minor comment about a `TODO`.

src/wavm/openmp.cpp
Shillaker: Yes good q, it seems to work for now but I will be digging into it more with the next phase of OpenMP work, so will leave it there as as note to self.

PR #400, opened 2021-03-30
Reclaiming memory from snapshots, sbrk, munmap and thread stacks

Shillaker:
Before these changes, memory-intensive applications would run out of memory because we were unable to reclaim it. This was because:

- Snapshots for spawning threads were created but not cleared away, so led to issues with applications doing lots of thread spawning
- `dlmalloc` had unmapping switched off (so wasn't even trying to deallocate memory)
- `dlmalloc` was using mmap-style allocation making it harder for us to reason about reclaiming memory (vs. sbrk-style)
- Thread stacks were created at arbitrary locations in the heap, breaking up the otherwise contiguously allocated regions (hence making it difficult to reclaim large chunks)

Changes:

- Change our fork of wasi-libc's dlmalloc to use sbrk _and_ allow deallocating memory
- Add locking to dlmalloc to keep it thread-safe
- Use our own `sbrk` implementation, which supports allocation and deallocation, but only allocates new memory regions when necessary, and reuses already allocated memory if possible.
- Add a thread stack pool which ensures all thread stacks live in a bunch at the bottom of the heap. This avoids having thread stacks interleaved with the rest of the heap memory, improving performance and making it easier to reclaim memory.
- New `threads` module to hold the state related to threading (including new `MutexManager` used by the pthread mutex implementation needed by dlmalloc).

With the amount of mutexes and locking going on, it's counterproductive to let OpenMP applications spawn arbitrary numbers of threads, as contention kills performance as we go above the number of cores on the machine. As a result, I've also added an OpenMP thread pooling approach where we don't kill threads once they're started (AFAICT this is what OpenMP does already). This is different to the old OpenMP thread pool where we created the pool up-front. 

I also updated the instructions for running Faasm outside of a container and using perf on wasm code. Finally I added the missing OpenMP functions from `faasm/cpp`, the only one we actually implement properly is `omp_get_wtime`.


docs/development.md
```diff
@@ -22,29 +22,29 @@ Docker compose config](docker-compose-dev.yml).
 We mount local checkouts of all the code into these containers, so first you'll
```
Shillaker, line 22: I don't know why it's saying this file has changed, when I run the diff locally it doesn't show up :thinking:

faasmcli/faasmcli/tasks/wast.py
```diff
@@ -19,7 +19,7 @@ def wast(ctx, user, func_name):
     """
     Generate .wast file for a given function
     """
-    func_dir = join(PROJ_ROOT, "wasm", user, func_name)
+    func_dir = join(WASM_DIR, user, func_name)
```
Shillaker, line 22: This was broken before

include/wamr/WAMRWasmModule.h
```diff
@@ -37,9 +37,7 @@ class WAMRWasmModule final : public WasmModule
     void tearDown();
 
     // ----- Memory management -----
-    uint32_t mmapMemory(uint32_t length) override;
-
-    uint32_t mmapPages(uint32_t pages) override;
```
Shillaker, line 42: We don't need the `mmapPages` function anywhere any more

include/wasm/WasmModule.h
```diff
@@ -23,13 +23,15 @@
 // Note: this is *not* controlling the size provisioned by the linker, that is
 // hard-coded in the build. This variable is just here for reference and must be
 // updated to match the value in the build.
-#define STACK_SIZE 4 * ONE_MB_BYTES
+#define STACK_SIZE (4 * ONE_MB_BYTES)
+#define THREAD_STACK_SIZE (2 * ONE_MB_BYTES)
+#define THREAD_STACK_POOL_SIZE 20
```
Shillaker, line 28: Threads are now handled at the generic `WasmModule` level, rather than only in the `WAVMWasmModule`

src/runner/CMakeLists.txt
```diff
@@ -25,3 +25,6 @@ target_link_libraries(codegen_func ${RUNNER_LIBS})
 add_executable(pool_runner pool_runner.cpp)
 target_link_libraries(pool_runner ${RUNNER_LIBS})
 
+# TEMP - for dev
+add_executable(covid_runner covid_runner.cpp)
+target_link_libraries(covid_runner ${RUNNER_LIBS})
```
Shillaker, line 30: I'll eventually delete the `covid_runner` but it's proving very useful in aggressively testing the OpenMP implementation.

src/wasm/WasmModule.cpp
```diff
@@ -260,6 +276,70 @@ uint32_t WasmModule::mapSharedStateMemory(
     return sharedMemWasmPtrs[segmentKey];
 }
 
+uint32_t WasmModule::getCurrentBrk()
+{
+    return currentBrk;
+}
+
+uint32_t WasmModule::createMemoryGuardRegion()
```
Shillaker, line 284: This used to live in the `WAVMWasmModule`

src/wasm/WasmModule.cpp
```diff
@@ -260,6 +276,70 @@ uint32_t WasmModule::mapSharedStateMemory(
...
+        throw std::runtime_error("Failed to create memory guard");
+    }
+
+    logger->debug(
+      "Created guard region {}-{}", wasmOffset, wasmOffset + regionSize);
+
+    return wasmOffset;
+}
+
+void WasmModule::createThreadStackPool()
```
Shillaker, line 309: This sets up our pool of thread stacks. It's called early on so they live at the bottom of the heap in a bunch.

src/wasm/WasmModule.cpp
```diff
@@ -260,6 +276,70 @@ uint32_t WasmModule::mapSharedStateMemory(
...
+}
+
+void WasmModule::returnThreadStack(uint32_t wasmPtr)
+{
+    auto logger = faabric::util::getLogger();
+    logger->debug("Returned thread stack {}", wasmPtr);
+    threadStacks.push_back(wasmPtr);
+}
+
+uint32_t WasmModule::allocateThreadStack()
```
Shillaker, line 330: This was the function that was used previously to provision thread stacks on the fly in arbitrary places in the heap. Now it just returns an offset to one of the existing stacks.

src/wavm/WAVMWasmModule.cpp
```diff
@@ -1003,86 +1008,171 @@ U32 WAVMWasmModule::mmapFile(U32 fd, U32 length)
...
+        logger->debug("MEM - munmapping top of memory by {}", nBytes);
+        shrinkMemory(nBytes);
+    } else {
+        logger->warn("MEM - unable to reclaim memory {} at {}", nBytes, offset);
+    }
+}
+
+U32 WAVMWasmModule::mmapMemory(U32 nBytes)
+{
+    return growMemory(nBytes);
```
Shillaker, line 1175: Although we still provide an `mmap` call, it's translated into growing/ shrinking linear memory.

src/wavm/memory.cpp
```diff
@@ -146,152 +137,46 @@ WAVM_DEFINE_INTRINSIC_FUNCTION(env,
...
-I32 s__brk(I32 addr)
-{
-    const std::shared_ptr<spdlog::logger>& logger = faabric::util::getLogger();
-    logger->debug("S - brk - {}", addr);
-
-    return _do_brk(addr);
-}
-
-I32 s__sbrk(I32 increment)
+WAVM_DEFINE_INTRINSIC_FUNCTION(env, "__sbrk", I32, __sbrk, I32 increment)
```
Shillaker, line 167: This is now the entrypoint used by `dlmalloc`

src/wavm/openmp.cpp
```diff
@@ -153,6 +154,20 @@ WAVM_DEFINE_INTRINSIC_FUNCTION(env,
     return ctx.threadNumber;
 }
 
+// ------------------------------------------------
+// TIMING
+// ------------------------------------------------
+
+WAVM_DEFINE_INTRINSIC_FUNCTION(env, "omp_get_wtime", F64, omp_get_wtime)
```
Shillaker, line 161: This was added in `faasm/cpp` and needed to be put here too. It's tested via a function in that repo.

src/wavm/syscalls.cpp
```diff
@@ -176,16 +176,12 @@ I32 executeSyscall(int syscallNumber,
             return s__rename(a, b);
         case 39:
             return s__mkdir(a, b);
-        case 45:
```
Shillaker, line 179: This whole `syscalls.cpp` file is the old way we used to do syscalls. It can be removed once the dust has settled on this current work.

faasmcli/faasmcli/tasks/dev.py
```diff
@@ -18,7 +18,9 @@
 
 
 @task
-def cmake(ctx, clean=False, build="Debug", native=False, perf=False):
+def cmake(
+    ctx, clean=False, build="Debug", native=False, perf=False, prof=False
+):
```
Shillaker, line 23: This was the change I mentioned for supporting local print debugging of performance.

src/wasm/WasmModule.cpp
```diff
@@ -36,6 +46,12 @@ size_t getNumberOfWasmPagesForBytes(uint32_t nBytes)
     return pageCount;
 }
 
+uint32_t roundUpToWasmPageAligned(uint32_t nBytes)
+{
+    size_t nPages = getNumberOfWasmPagesForBytes(nBytes);
+    return (uint32_t)(nPages * WASM_BYTES_PER_PAGE);
```
Shillaker, line 52: Could be more efficient but it's readable and only used in a couple of places atm.

src/wasm/WasmModule.cpp
```diff
@@ -60,43 +76,58 @@ wasm::WasmEnvironment& WasmModule::getWasmEnvironment()
     return wasmEnvironment;
 }
 
-std::string WasmModule::snapshot()
+std::string WasmModule::snapshot(bool locallyRestorable)
```
Shillaker, line 79: This `locallyRestorable` parameter is passed through to Faabric to signify whether a snapshot is intended purely for remote process/ thread spawning

src/wavm/WAVMWasmModule.cpp
```diff
@@ -1003,86 +1010,176 @@ U32 WAVMWasmModule::mmapFile(U32 fd, U32 length)
...
+    size_t newBytes = oldBytes + nBytes;
+
+    uint32_t oldBrk = currentBrk;
+    uint32_t newBrk = currentBrk + nBytes;
 
-    if (pages == 0) {
-        throw std::runtime_error("Requesting mapping of zero pages");
+    if (!isWasmPageAligned(newBrk)) {
+        logger->error("Growing memory by {} is not wasm page aligned", nBytes);
+        throw std::runtime_error("Non-wasm-page-aligned memory growth");
```
Shillaker, line 1032: We check wasm page alignment in all these functions because they should only ever be called behind the scenes by libc and malloc. Where we call them internally, we have to make sure we round up to the nearest wasm page.

src/wavm/openmp.cpp
```diff
@@ -945,5 +980,37 @@ WAVM_DEFINE_INTRINSIC_FUNCTION(env,
     OMP_FUNC_ARGS("omp_set_default_device {} (ignored)", defaultDeviceNumber);
 }
 
+// ----------------------------------------------
+// ATOMICS
+// ----------------------------------------------
+
+WAVM_DEFINE_INTRINSIC_FUNCTION(env,
```
Shillaker, line 987: Again, for compatibility with the latest `faasm/cpp`

include/wasm/WasmModule.h
```diff
@@ -86,12 +90,20 @@ class WasmModule
...
 
-    virtual uint32_t mmapPages(uint32_t pages);
+    virtual uint32_t mmapMemory(uint32_t nBytes);
 
     virtual uint32_t mmapFile(uint32_t fp, uint32_t length);
 
+    virtual void unmapMemory(uint32_t offset, uint32_t nBytes);
+
+    uint32_t createMemoryGuardRegion();
+
```
Shillaker, line 106: Hopefully the `WasmModule` interface can become more and more generic so we can share more functionality between WAMR and WAVM (and VMs we add in future...)

include/threads/MutexManager.h
```diff
@@ -0,0 +1,29 @@
...
+  public:
+    void createMutex(int mutexId);
+
+    void lockMutex(int mutexId);
+
+    void unlockMutex(int mutexId);
+
+    bool tryLockMutex(int mutexId);
+
+    void destroyMutex(int mutexId);
```
Shillaker, line 20: This is a new class to handle shared mutexes between modules. It's really just a map and wrapper around the existing `std::thread` but it means the implementation can be shared by all types of `WasmModule` and avoids nesting this complexity in the intrinsic threading functions.

include/wasm/WasmModule.h
```diff
@@ -121,8 +142,8 @@ class WasmModule
     int stdoutMemFd;
     ssize_t stdoutSize;
 
-    std::mutex moduleMemoryMutex;
```
Shillaker, line 124: Now that we use a thread-safe dlmalloc we don't need to do our own locking on the module memory.

src/wavm/WAVMWasmModule.cpp
```diff
@@ -565,9 +571,11 @@ Runtime::Instance* WAVMWasmModule::createModuleInstance(
           boundUser, boundFunction, sharedModulePath);
 
         // Provision the memory for the new module plus two guard regions
-        createMemoryGuardRegion();
-        Uptr newMemory = mmapPages(DYNAMIC_MODULE_MEMORY_PAGES);
-        createMemoryGuardRegion();
+        uint32_t memSize = DYNAMIC_MODULE_MEMORY_SIZE + (2 * GUARD_REGION_SIZE);
+        Uptr newMemory = growMemory(memSize);
+        createMemoryGuardRegion(newMemory);
+        createMemoryGuardRegion(newMemory + DYNAMIC_MODULE_MEMORY_SIZE +
```
Shillaker, line 577: Note that before, creating a stack with memory guards either side performed three memory mappings, now we do one, and apply the read-only protection on the guard pages afterwards.

src/wavm/threads.cpp
```diff
@@ -286,83 +286,123 @@ I32 s__futex(I32 uaddrPtr,
...
+// support thread-safe libc operations
+// --------------------------
 
 WAVM_DEFINE_INTRINSIC_FUNCTION(env,
                                "pthread_mutex_init",
                                I32,
                                pthread_mutex_init,
-                               I32 a,
-                               I32 b)
+                               I32 mx,
```
Shillaker, line 298: Note that we just use the integer wasm offset for the given mutex to identify it.

src/wavm/threads.cpp
```diff
@@ -286,83 +286,123 @@ I32 s__futex(I32 uaddrPtr,
...
+                               s__pthread_mutex_trylock,
+                               I32 mx)
 {
-    // faabric::util::getLogger()->trace("S - pthread_cond_init {} {}", a, b);
+    faabric::util::getLogger()->debug("S - pthread_mutex_trylock {}", mx);
+    bool success = threads::getMutexManager().tryLockMutex(mx);
+    if (success) {
+        return 0;
+    } else {
+        return EBUSY;
```
Shillaker, line 328: `pthread_mutex_trylock` has a richer set of error codes, but in our case we can assume it's just busy.

include/threads/ThreadState.h
```diff
@@ -57,6 +56,37 @@ class Level
     std::condition_variable nowaitCv;
 };
 
+class OpenMPTask
```
Shillaker, line 59: This is the task that's put onto the queue for the thread pool.

include/threads/ThreadState.h
```diff
@@ -57,6 +56,37 @@ class Level
...
+{
+  public:
+    OpenMPTask(faabric::Message* parentMsgIn,
+               faabric::Message& msgIn,
+               std::shared_ptr<threads::Level> nextLevelIn,
+               int threadIdxIn)
```
Shillaker, line 65: Should be using smart pointers for the `parentMsg` and `msg` but that would required even more changes in this already bloated PR.

include/wasm/WasmModule.h
```diff
@@ -20,16 +23,19 @@
 
 #define WASM_BYTES_PER_PAGE 65536
 
+#define OPENMP_QUEUE_TIMEOUT_MS 10000
```
Shillaker, line 26: This should only be hit when things have failed, it's to avoid an indefinite hanging when things break.

include/wasm/WasmModule.h
```diff
@@ -121,7 +159,15 @@ class WasmModule
     int stdoutMemFd;
     ssize_t stdoutSize;
 
-    std::mutex moduleMemoryMutex;
+    faabric::util::Queue<threads::OpenMPTask> openMPTaskQueue;
+    faabric::util::Queue<int32_t> openMPResultQueue;
+    std::vector<std::thread> openMPThreads;
+    std::mutex threadsMutex;
+    std::mutex threadStacksMutex;
```
Shillaker, line 166: These extra instance variables associated with the wasm module are unfortunate. Executing OpenMP threads is quite involved, accessing other internals of the wasm module class, so this was the easiest way to do it for now. I'd like to try and factor this out into a separate thread pool class eventually.

src/runner/covid_runner.cpp
```diff
@@ -0,0 +1,75 @@
...
+#include <faabric/util/config.h>
+#include <faabric/util/func.h>
+#include <faaslet/FaasletPool.h>
+
+#include <module_cache/WasmModuleCache.h>
+#include <wamr/WAMRWasmModule.h>
+
+#include <faabric/util/files.h>
+
+int main(int argc, char* argv[])
```
Shillaker, line 16: I'll delete the covid runner once we're done with developing this use-case, but it's very useful for now.

include/threads/ThreadState.h
```diff
@@ -57,11 +56,44 @@ class Level
...
+{
+  public:
+    PthreadTask(faabric::Message* parentMsgIn,
+                std::shared_ptr<faabric::Message> msgIn)
+      : parentMsg(parentMsgIn)
+      , msg(msgIn)
+    {}
+
+    bool isShutdown = false;
+    faabric::Message* parentMsg;
```
Shillaker, line 69: Unfortunately the parent message comes from `getExecutingCall` which is used all over the place and returns a raw pointer. We know that the lifetime of this message exceeds the threaded call though, so it's ok for now.

include/conf/FaasmConfig.h
```diff
@@ -0,0 +1,23 @@
+#pragma once
```
Shillaker, line 1: I couldn't bring myself to add another Faasm-specific parameter to the Faabric `SystemConfig`, so this is the beginnings of splitting out the Faasm-specific config from there.

src/wavm/WAVMWasmModule.cpp
```diff
@@ -851,11 +858,12 @@ bool WAVMWasmModule::execute(faabric::Message& msg, bool forceNoop)
...
 
     // Set up OMP
-    setUpOpenMPContext(msg);
+    threads::setUpOpenMPContext(msg);
 
     // Executes OMP fork message if necessary
     if (msg.ompdepth() > 0) {
-        return executeAsOMPThread(msg);
+        // TODO - reimplement this call to execute a remote OpenMP thread
+        throw std::runtime_error("Not yet implemented remote OpenMP");
```
Shillaker, line 866: Turns out this branch wasn't getting executed at all in tests any more. I'll resurrect as part of the next phase on cross-host OpenMP

src/wasm/WasmModule.cpp
```diff
@@ -60,43 +94,59 @@ wasm::WasmEnvironment& WasmModule::getWasmEnvironment()
...
 {
+    PROF_START(wasmSnapshot)
```
csegarragonz, line 99: I am wondering, should this `PROF_START` calls be merged in? I know they becom `no-op`s but I'd say these were useful for your debugging, but not to someone else who runs with trace?

csegarragonz (changes requested):
Some comments, all very superificial though.

src/threads/MutexManager.cpp
```diff
@@ -0,0 +1,32 @@
...
+
+namespace threads {
+
+void MutexManager::createMutex(int mutexId)
+{
+    mutexes[mutexId];
+}
+
+void MutexManager::lockMutex(int mutexId)
+{
```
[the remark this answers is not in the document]
Shillaker: I know what you're saying and yes, we're actually deliberately using the default initialisation here. Some things call `lockMutex` without ever initialising the mutex first, hence we have to support creating the locks on the fly. I could put a comment in this file to signal that this is the expected behaviour.

src/wamr/pthread.cpp
```diff
@@ -1,27 +1,66 @@
...
-// ------------------------------------------
-// NOTE: these functions are just here to stub out bits that are missing
-// from the existing WAMR pthread implementation. This can be found here:
-// https://github.com/bytecodealliance/wasm-micro-runtime/tree/main/core/iwasm/libraries/lib-pthread
-// ------------------------------------------
+static int32_t pthread_mutex_init_wrapper(wasm_exec_env_t exec_env,
+                                          int32_t a,
+                                          int32_t b)
+{
+    faabric::util::getLogger()->debug("S - pthread_mutex_init {} {}", a, b);
```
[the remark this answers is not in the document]
Shillaker: This is the WAMR implementation of these functions, and we don't support multi-threading with WAMR yet, therefore it's not possible to execute multi-threaded applications that will invoke these. However, we can't throw an exception as they will always be called by malloc/ free now that we use thread-safe dlmalloc. I guess we could add the implementations now, but there is a lot more work to be done to make the WAMR implementation support multi-threading. I think a comment here to explain this might be better.

src/wasm/WasmModule.cpp
Shillaker: I would agree for some of the places I've put them in this PR should be removed as they're overkill, but I actually think having `PROF_START` and `PROF_END` around important, time-consuming stuff like snapshot and restore is very useful. Whenever someone switches this stuff on, they are going to be looking at a trace log, which will be huge to start with, so I wouldn't worry about a few extra lines. More importantly, they will be looking to explain what's taking time in their application, and this could be a source of latency that they weren't considering, so might pop out with something useful.

src/wavm/threads.cpp
```diff
@@ -187,14 +136,11 @@ WAVM_DEFINE_INTRINSIC_FUNCTION(env,
 
     int returnValue;
 
-    if (localThreads.count(pthreadPtr) > 0) {
-        // Get the local thread and remove it from the local map
-        Platform::Thread* thread = localThreads[pthreadPtr];
-        localThreads.erase(pthreadPtr);
-
-        // Join it
+    if (localThreadFutures.count(pthreadPtr) > 0) {
```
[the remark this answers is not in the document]
Shillaker: I realise it's more efficient but would argue that the readability trade-off is worth it. I personally find comparing the output of `find` to `end` isn't particularly easy to skim over, then when it comes to erasing you have to look back a couple of lines to find what the iterator is referring to. If we're going to nitpick over efficiency of different ways to skin the cat there are probably more obvious places to start in this codebase...

csegarragonz (approved):

PR #402, opened 2021-04-01
Updates for perf and building outside of containers

Shillaker:
Instructions were out of date and I added some clarity around the use of perf.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -39,7 +39,8 @@ set(FAASM_INCLUDE_DIR ${CMAKE_CURRENT_LIST_DIR}/include)
 set(FAASM_NATIVE_INSTALL "/usr/local/faasm/native")
 
 # CPP REST SDK
-find_package(cpprestsdk REQUIRED)
+set(cpprestsdk_DIR /usr/lib/${CMAKE_LIBRARY_ARCHITECTURE}/cmake/)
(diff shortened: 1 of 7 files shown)


CMakeLists.txt
```diff
@@ -39,7 +39,8 @@ set(FAASM_INCLUDE_DIR ${CMAKE_CURRENT_LIST_DIR}/include)
 set(FAASM_NATIVE_INSTALL "/usr/local/faasm/native")
 
 # CPP REST SDK
-find_package(cpprestsdk REQUIRED)
+set(cpprestsdk_DIR /usr/lib/${CMAKE_LIBRARY_ARCHITECTURE}/cmake/)
+find_package(cpprestsdk REQUIRED NAMES cpprestsdk cpprest)
```
Shillaker, line 43: This fixes for use with `apt` on some systems

bin/build_llvm_perf.sh
```diff
@@ -2,33 +2,35 @@
 set -e
 set -x
 
-CODE_DIR=/usr/local/code/
```
Shillaker, line 5: Btw, this script is for installing a special build of LLVM that allows us to use perf with the JIT modules. This isn't included in the default LLVM distros.

csegarragonz (approved):

PR #405, opened 2021-04-06
Fixes for LAMMPS - Rebuilt

csegarragonz:
UPDATE: for some reason, #403 had cached some funny sbumodule commit references that were making the tests fail. I've cherry-picked the changes to this new PR, and will delete the other one.

This is a companion PR to faasm/faabric#67 and faasm/experiment-lammps#8. I also update the tests in faasm/cpp#39.

It is important to acknowledge that, even though faasm/cpp#39 may get merged in, `faasm` tracks faasm/cpp#34 -> faasm/cpp#39 instead of the most recent `master` there. Hence we must not mess the commit references.

Additionally, I have also upgraded the logging for MPI, by relying on the enhanced logger. Now, there is a per-rank logger.

### DIFF
--- a/src/wavm/mpi.cpp
+++ b/src/wavm/mpi.cpp
@@ -14,6 +14,7 @@ using namespace WAVM;
 
 namespace wasm {
 static thread_local faabric::scheduler::MpiContext executingContext;
+static thread_local std::shared_ptr<spdlog::logger> mpiLogger = nullptr;
 
 bool isInPlace(U8 wasmPtr)
 {
@@ -64,6 +65,16 @@ class ContextWrapper
         }
 
         return true;
+    }
+
+    std::shared_ptr<spdlog::logger> getMpiLogger()
+    {
+        if (mpiLogger == nullptr) {
+            mpiLogger = faabric::util::getLogger(
+              fmt::format("MPI-{}", executingContext.getRank()));
+        }
+
+        return mpiLogger;
     }
 
     faabric_datatype_t* getFaasmDataType(I32 wasmPtr)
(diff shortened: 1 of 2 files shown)


src/wavm/mpi.cpp
```diff
@@ -594,11 +610,12 @@ WAVM_DEFINE_INTRINSIC_FUNCTION(env,
                                I32 requestPtrPtr,
                                I32 status)
 {
-    faabric::util::getLogger()->debug(
-      "S - MPI_Wait {} {}", requestPtrPtr, status);
-
     ContextWrapper ctx;
+
+    // ctx.getMpiLogger()->debug("S - MPI_Wait {} {}", requestPtrPtr, status);
```
Shillaker, line 615: Can we lose this commented line in some future PR? Doesn't need to be done now

src/wavm/mpi.cpp
```diff
@@ -647,14 +666,16 @@ WAVM_DEFINE_INTRINSIC_FUNCTION(env,
 
 WAVM_DEFINE_INTRINSIC_FUNCTION(env, "MPI_Abort", I32, MPI_Abort, I32 a, I32 b)
 {
-    faabric::util::getLogger()->debug("S - MPI_Abort {} {}", a, b);
+    ContextWrapper ctx;
+    ctx.getMpiLogger()->debug("S - MPI_Abort {} {}", a, b);
     //        return terminateMpi();
```
Shillaker, line 671: Not part of this change but we should probably think about what terminating MPI means, whether we can clean up anything.

Shillaker (approved):

src/wavm/mpi.cpp
csegarragonz: Good catch, thanks.

PR #406, opened 2021-04-07
Docs update + Profiling flag

csegarragonz:
Very minor tweaks I've done whilst trying to build and run binaries outside the container environment.


.env
```diff
@@ -1,4 +1,5 @@
 FAASM_VERSION=0.5.14
+FAASM_CLI_IMAGE=faasm/cli:0.5.14
```
csegarragonz, line 2: Following the `Quick Start` running `docker-compose up -d` without this variable in the `.env` file would fail.

PR #409, opened 2021-04-21
More OpenMP performance

Shillaker:
Changes:

- Change OpenMP thread pooling to use consistent OpenMP thread numbers for each thread in the pool. This ensure the pool is only ever as big as it needs to be, avoiding needless contention on the queue and initialising threads unnecessarily.
- Reuse WAVM `Context`s between OpenMP tasks. Removes overhead associated with each.
- Update local cluster dev environment and instructions.


bin/cluster_dev.sh
```diff
@@ -23,15 +23,27 @@ fi
...
-    faasm-cli
+
+if [[ "$1" == "faasm" ]]; then
+    CLI_CONTAINER="faasm-cli"
+elif [[ "$1" == "cpp" ]]; then
+    CLI_CONTAINER="cpp"
+else
+    echo "Not starting a CLI container"
+    exit 0
+fi
```
Shillaker, line 40: Acts a bit like `cli.sh` for the development cluster. Although they're similar, it's good to keep the two separate as the dev cluster is much more like a "real" cluster while the development set-up is very inter-connected with a lot of mounts.

faasmcli/faasmcli/tasks/redis.py
```diff
@@ -1,15 +1,16 @@
...
 from faasmcli.util.env import PROJ_ROOT
 
 
 def _do_redis_command(sub_cmd, local, docker, knative):
+    redis_host = environ.get("REDIS_QUEUE_HOST", "redis")
     if local:
         cmd = ["redis-cli", sub_cmd]
     elif docker:
-        cmd = ["docker-compose", "exec", "redis-queue", "redis-cli", sub_cmd]
+        cmd = ["redis-cli", "-h", redis_host, sub_cmd]
```
Shillaker, line 13: This is now run inside the container rather than outside.

include/threads/ThreadState.h
```diff
@@ -76,17 +76,14 @@ class OpenMPTask
     faabric::Message* parentMsg;
     std::shared_ptr<faabric::Message> msg;
     std::shared_ptr<threads::Level> nextLevel;
-    int threadIdx;
```
Shillaker, line 79: There were too many things labelled "threadIdx" and this was no longer used.

src/wavm/WAVMWasmModule.cpp
```diff
@@ -376,6 +382,20 @@ void WAVMWasmModule::addModuleToGOT(IR::Module& mod, bool isMainModule)
...
+  const std::vector<WAVM::IR::UntaggedValue>& arguments,
+  WAVM::IR::UntaggedValue& result)
+{
+    // Note the need to set the currently executing module
+    setExecutingModule(this);
+
+    const IR::FunctionType funcType = Runtime::getFunctionType(func);
+
+    Runtime::invokeFunction(ctx, func, funcType, arguments.data(), &result);
+}
```
Shillaker, line 397: We didn't have a way to execute a function with a different context before

src/wavm/openmp.cpp
```diff
@@ -819,6 +815,13 @@ WAVM_DEFINE_INTRINSIC_FUNCTION(env,
 
 int reduceFinished()
 {
+    auto& ctx = threads::getOpenMPContext();
+
+    // Notify the master thread that we've done our reduction
+    if (ctx.threadNumber != 0) {
+        ctx.level->masterWait(ctx.threadNumber);
+    }
```
Shillaker, line 823: Turns out this locking was necessary, but actually doesn't cause much overhead

include/wasm/WasmModule.h
```diff
@@ -42,6 +42,14 @@
 #define WASM_CTORS_FUNC_NAME "__wasm_call_ctors"
 #define ENTRY_FUNC_NAME "_start"
 
+typedef std::pair<std::promise<int32_t>, threads::OpenMPTask> OpenMPTaskPair;
+typedef std::pair<std::promise<int32_t>, threads::PthreadTask> PthreadTaskPair;
+
+typedef faabric::util::Queue<OpenMPTaskPair> OpenMPQueue;
+typedef faabric::util::Queue<PthreadTaskPair> PthreadQueue;
+
+typedef std::unordered_map<int, OpenMPQueue> OpenMPQueueMap;
```
Shillaker, line 51: Hide some of the grimness :smile:

include/wasm/WasmModule.h
```diff
@@ -159,12 +168,10 @@ class WasmModule
...
 
-    faabric::util::Queue<std::pair<std::promise<int32_t>, threads::OpenMPTask>>
-      openMPTaskQueue;
-    faabric::util::Queue<std::pair<std::promise<int32_t>, threads::PthreadTask>>
-      pthreadTaskQueue;
+    OpenMPQueueMap openMPTaskQueueMap;
+    PthreadQueue pthreadTaskQueue;
 
-    std::vector<std::thread> openMPThreads;
+    std::unordered_map<int, std::thread> openMPThreads;
```
Shillaker, line 174: We care about which thread is which, and may not need all of them so a map seemed to make sense. It could also be an array of pointers, which is what we use for the thread contexts in the `WAVMModule`. Given that this works and doesn't seem to be a performance issue I'm inclined to leave that for later.

src/wasm/WasmModule.cpp
```diff
@@ -430,69 +426,77 @@ std::future<int32_t> WasmModule::executeOpenMPTask(threads::OpenMPTask t)
...
-
-                    taskPair.first.set_value(
-                      executeAsOMPThread(stackTop, taskPair.second.msg));
-
-                    // Caller has to notify scheduler when finished
-                    // executing a thread locally
-                    sch.notifyCallFinished(*taskPair.second.msg);
-                }
-            });
+            openMPThreads.insert(
```
Shillaker, line 455: Mostly just indentation here

src/wavm/WAVMWasmModule.cpp
```diff
@@ -969,21 +989,23 @@ int32_t WAVMWasmModule::executeAsPthread(uint32_t stackTop,
...
-    };
+    Runtime::ContextRuntimeData* contextRuntimeData =
+      getContextRuntimeData(executionContext);
+
+    // Set up the context
+    Runtime::Context* threadContext =
+      createThreadContext(stackTop, contextRuntimeData);
 
     // Record the return value
-    I64 returnValue = executeThreadLocally(spec);
```
Shillaker, line 980: The `executeThreadLocally` function was no longer useful, hence the slight refactor of both the OpenMP and pthread bits.

src/wasm/WasmModule.cpp
```diff
@@ -430,69 +426,77 @@ std::future<int32_t> WasmModule::executeOpenMPTask(threads::OpenMPTask t)
...
 
-    // Check if we can add a thread if we're not shutting down
-    if (!t.isShutdown && (openMPThreads.size() < threadPoolSize)) {
+    // We have a fixed pool of threads, we have to share the work evenly over
+    // them, but we always want to send the same OMP thread number to the same
+    // thread. However, if we don't have enough threads in the pool, we want to
+    // cycle round again, and if we're at a nested level, we want to spread the
+    // work to other threads so we don't get nested ones blocking each other.
+    int threadPoolIdx =
+      (t.nextLevel->depth - 1 + t.msg->ompthreadnum()) % threadPoolSize;
```
csegarragonz, line 436: Just to double-check, can `t.nextLevel->depth` ever be 0? This is, can:
```cpp
t.nextLevel->depth - 1 + t.msg->ompthreadnum()
```
ever be -1? I guess this would yield undefined behaviour.

src/wasm/WasmModule.cpp
csegarragonz: Is there a reason for switching from `emplace`-ing to `insert`-ing? I would personally always go for the former as it can sometimes be benefficial, and won't harm.

src/wavm/WAVMWasmModule.cpp
```diff
@@ -88,6 +88,9 @@ WAVMWasmModule::WAVMWasmModule()
 {
     stdoutMemFd = 0;
     stdoutSize = 0;
+
+    // Prepare OpenMP context cache
+    openMPContexts = std::vector<Runtime::Context*>(threadPoolSize, nullptr);
```
csegarragonz, line 93: I am not familiar with the `Runtime::Context` class, but do you think we could beneffit from using a smart pointer here?

src/wavm/openmp.cpp
```diff
@@ -512,7 +508,7 @@ WAVM_DEFINE_INTRINSIC_FUNCTION(env,
         setUpOpenMPContext(0, nextLevel);
 
         // Execute the task
-        logger->debug("Executing master OMP thread");
+        logger->debug("OpenMP 0: executing OMP thread 0 (master)");
```
csegarragonz, line 511: Just thinking out loud here, but you could have a `thread local` logger, and use something like:
```cpp
auto logger = faabric::util::getLogger(fmt::format("OMP-{}", threadIdx));
```
This is not a request for changes, rather a suggestion if you find it useful.

csegarragonz (approved):
LGTM. Some comments.

src/wavm/WAVMWasmModule.cpp
Shillaker: Unfortunately not as it's not created by us, WAVM returns us a raw pointer.

src/wavm/openmp.cpp
Shillaker: Hmm, I wonder what's better in terms of contention, a single logger used by all threads or a logger per thread? I would go with whatever is more performant, my gut being a single logger used by all threads, but I may be wrong.

src/wasm/WasmModule.cpp
Shillaker: Yes it's a good point. I'm 90% sure the depth of the lowest level is `1`, but I will definitely check, and perhaps throw an exception if the depth is less than 1.

src/wavm/openmp.cpp
csegarragonz: Well, you are probably right. But one could argue that when measuring performance logging will be turned off.

PR #410, opened 2021-04-22
Shared file performance

Shillaker:
Attempts to improve performance of shared file access.


Shillaker:
This will be dealt with in https://github.com/faasm/faasm/pull/432

PR #411, opened 2021-04-22
Update to Faabric batch scheduling

Shillaker:
This PR has totally spiralled, it's exposed a few concurrency issues that have led to some whack-a-mole rewriting and refactoring. In general I've tried to remove complexity wherever possible, and add assertions and checks to catch issues.

Summary of changes:

- Update Faasm to the simpler executor model introduced in Faabric: https://github.com/faasm/faabric/pull/83 
- Make a better distinction between zygotes, cached modules and snapshots
- Isolate WAVM module caching inside the `wavm` module (as this is very specific). Previously this was done through a separate `module_cache` module which made the logic hard to follow. 
- Merge the `ir_cache` module into the `wavm` module as it's also WAVM-specific (and maybe a candidate for complete removal?)
- Remove `simple_runner` and replace with `func_runner` as now that the executor pooling has been removed, they are essentially equivalent.
- Avoid use of destructors unless absolutely necessary to release resources, they had become used for general tidy-up and introduced some nasty errors when they went wrong.
- Encapsulate the indexing of `NetworkNamespace` objects inside the class, rather than having wasm modules manage this
- Avoid `thread_local` global variables where possible, as thread-locality isn't always a guarantee of separation between modules/ functions. Where they do still exist (e.g. with `get`/`setExecutingModule`), I've added some relatively strict `assert`s.

There are some OpenMP-specific changes too:

- To pass control of distributed threading to Faabric, the OpenMP thread context now needs to be serialised and sent with the batch execution requests (rather than managed by the wasm module). 
- The synchronisation operations such as barriers, mutexes and critical sections need to support look-up by the `Level` id, to allow distributed threads to find them (without relying on them being available in-memory)


docker-compose.yml
```diff
@@ -36,6 +36,7 @@ services:
     privileged: yes
     volumes:
       - ./dev/faasm/build/:${FAASM_BUILD_MOUNT}
+      - ./dev/container/shared_store/:/usr/local/faasm/shared_store/
```
Shillaker, line 39: Revert this

CMakeLists.txt
```diff
@@ -175,8 +175,6 @@ include_directories(
 # Faasm runtime
 add_subdirectory(src/conf)
 add_subdirectory(src/faaslet)
-add_subdirectory(src/ir_cache)
-add_subdirectory(src/module_cache)
```
Shillaker, line 179: Both merged into `src/wavm`

include/threads/ThreadState.h
```diff
@@ -10,6 +11,20 @@
...
+{
+  public:
+    int32_t depth;
+    int32_t effectiveDepth;
+    int32_t maxActiveLevels;
+    int32_t nThreads;
+
+    uint32_t nSharedVars = 0;
+    uint32_t* sharedVars;
+};
```
Shillaker, line 24: The `SerializedLevel` is the context that gets sent with remote batch OpenMP requests

include/wasm/WasmModule.h
```diff
@@ -42,16 +41,14 @@
 #define WASM_CTORS_FUNC_NAME "__wasm_call_ctors"
 #define ENTRY_FUNC_NAME "_start"
 
-typedef std::pair<std::promise<int32_t>, threads::OpenMPTask> OpenMPTaskPair;
-typedef std::pair<std::promise<int32_t>, threads::PthreadTask> PthreadTaskPair;
-
-typedef faabric::util::Queue<OpenMPTaskPair> OpenMPQueue;
-typedef faabric::util::Queue<PthreadTaskPair> PthreadQueue;
-
-typedef std::unordered_map<int, OpenMPQueue> OpenMPQueueMap;
```
Shillaker, line 51: Note that all the thread pooling is handled by Faabric now

docs/debugging.md
```diff
@@ -6,8 +6,8 @@ issues.
 
 ## Backtrace
 
-If you're seeing a failure nested somewhere deep in an application but can't 
-work out how it's getting there, you can use the `__faasm_backtrace` function 
+If you're seeing a failure nested somewhere deep in an application but can't
+work out how it's getting there, you can use the `__faasm_backtrace` function
```
Shillaker, line 10: You'll have to ignore the changes to the docs files as it's almost all whitespace trimming apart from swapping `simple_runner` for `func_runner`

include/threads/ThreadState.h
```diff
@@ -87,16 +108,10 @@ class OpenMPTask
     {}
 };
 
-class OpenMPContext
-{
-  public:
-    int threadNumber = -1;
-    std::shared_ptr<Level> level = nullptr;
-};
```
Shillaker, line 95: There was no need for this wrapper object, the only thing that's actually shared is the `Level`

faasmcli/faasmcli/tasks/flame.py
```diff
@@ -12,12 +12,12 @@
 
 
 @task(default=True)
-def general(ctx, user, func, reps=10, cmd=None, data=None, reverse=False):
+def general(ctx, user, func, cmd=None, data=None, reverse=False):
```
Shillaker, line 15: Including multiple repetitions in the runner scripts was only really useful for profiling, and added needless complexity otherwise.

include/faaslet/Faaslet.h
```diff
@@ -1,41 +1,50 @@
...
 
+  protected:
     void postFinish() override;
 
+    void restore(const faabric::Message& call) override;
+
   private:
-    int isolationIdx;
-    std::unique_ptr<isolation::NetworkNamespace> ns;
+    std::shared_ptr<isolation::NetworkNamespace> ns;
```
Shillaker, line 36: The changes in the `Faaslet` interface are just to conform with Faabric.

include/wavm/WAVMWasmModule.h
```diff
@@ -114,6 +117,9 @@ class WAVMWasmModule final
       uint32_t stackTop,
       WAVM::Runtime::ContextRuntimeData* contextRuntimeData);
 
+    std::unordered_map<int32_t, uint32_t> chainedThreads;
```
Shillaker, line 120: This is a map of wasm offsets to message IDs, so we can't use an array unfortunately.

tests/utils/worker_utils.cpp
```diff
@@ -1,15 +1,17 @@
+#include "faabric/util/config.h"
```
Shillaker, line 1: I've tried to make this file more high-level and do less interrogation of the Faabric internals, hence there's a lot of red.

tests/test/wasm/test_wasm.cpp
```diff
@@ -81,106 +90,66 @@ TEST_CASE("Test binding", "[wasm]")
...
-    SECTION("Bound but not executed module") { module.bindToFunction(call); }
-    SECTION("Executed module")
-    {
-        module.bindToFunction(call);
-        module.execute(call);
-    }
-
-    bool success = module.tearDown();
-    REQUIRE(success);
-}
```
Shillaker, line 180: WAVM GC will now fail an assertion if it fails. This should never happen unless we change the code in a way that makes it not possible to GC, in which case we should fix it.

tests/test/wasm/test_func_ptr.cpp
```diff
@@ -1,36 +0,0 @@
...
-    // Restore from the zygote and execute the function (expect it to succeed)
-    wasm::WAVMWasmModule moduleB;
-    moduleB.bindToFunction(call);
-
-    moduleB.restore(stateKey);
-    bool success = moduleB.execute(call);
-    REQUIRE(success);
-
-    // Return code will be equal to the failure case of the function
-    REQUIRE(call.returnvalue() == 101);
```
Shillaker, line 34: This test is really checking the internals of the WAVM wasm module, and so was very brittle. The ability to execute functions from pointers is implicitly tested all over the place.

tests/test/wasm/test_cloning.cpp
```diff
@@ -18,22 +21,6 @@ void convertMsgToPython(faabric::Message& msg)
     msg.set_ispython(true);
 }
 
-TEST_CASE("Test cloning empty modules doesn't break", "[wasm]")
```
Shillaker, line 21: This will now break as we should never be doing it.

tests/test/faaslet/test_threads.cpp
```diff
@@ -53,14 +32,4 @@ TEST_CASE("Run thread checks locally", "[threads]")
...
-
-TEST_CASE("Run thread checks with chaining", "[threads]")
-{
-    runTestDistributed("threads_check");
-}
-
-TEST_CASE("Run distributed threading check", "[threads]")
-{
-    runTestDistributed("threads_dist");
-}
```
Shillaker, line 65: These two functions weren't really testing distributed execution, they were just doing a _lot_ of fiddling with internals to fake up what we think distributed execution will do. Instead I'll leave this to another PR on doing "proper" distributed testing with multiple containers.

src/wavm/threads.cpp
```diff
@@ -20,10 +20,6 @@
 using namespace WAVM;
 
 namespace wasm {
-// Map of tid to message ID for chained calls
-static thread_local std::unordered_map<I32, unsigned int> chainedThreads;
-static thread_local std::unordered_map<I32, std::future<int32_t>>
-  localThreadFutures;
```
Shillaker, line 26: These are really specific to the module and not any given thread.

src/wavm/openmp.cpp
```diff
@@ -388,17 +377,18 @@ WAVM_DEFINE_INTRINSIC_FUNCTION(env,
     Runtime::Memory* memoryPtr = parentModule->defaultMemory;
     faabric::Message* parentCall = getExecutingCall();
 
-    // Set up number of threads for next level
```
Shillaker, line 391: Changes from here on are mostly to do with refactoring to the new Faabric set-up where we don't need to worry about spawning our own tasks etc.

src/wavm/WAVMWasmModule.cpp
```diff
@@ -250,26 +234,13 @@ bool WAVMWasmModule::tearDown()
...
-    if (!compartmentCleared) {
-        logger->warn("Failed GC for compartment");
-    } else {
-        logger->debug("Successful GC for compartment");
+    if (compartment != nullptr) {
+        // Release build complains that this is unused as the assertion is
+        // removed
+        [[maybe_unused]] bool compartmentCleared =
+          Runtime::tryCollectCompartment(std::move(compartment));
+        assert(compartmentCleared);
```
Shillaker, line 242: This `[[maybe_unused]]` is to fix an error when doing the release build, where it complains that `compartmentCleared` is unused (because the `assert` is no-oped)

src/wavm/WAVMWasmModule.cpp
```diff
@@ -210,37 +210,21 @@ void WAVMWasmModule::clone(const WAVMWasmModule& other)
...
-    // --- Faasm stuff ---
-    // Shared mem
-    sharedMemWasmPtrs.clear();
-
-    // Dynamic modules
-    globalOffsetTableMap.clear();
-    globalOffsetMemoryMap.clear();
-    missingGlobalOffsetEntries.clear();
-
-    dynamicPathToHandleMap.clear();
```
Shillaker, line 231: Resetting normal instance variables will be handled by the default destructor.

src/wavm/WAVMModuleCache.cpp
```diff
@@ -0,0 +1,61 @@
+#include <wavm/WAVMWasmModule.h>
```
Shillaker, line 1: This whole `WAVMModuleCache` class used to be the `WasmModuleCache` and have a lot more unnecessary complexity.

src/wasm/WasmModule.cpp
```diff
@@ -318,193 +362,64 @@ uint32_t WasmModule::getCurrentBrk()
...
-    pthreadTaskQueue.enqueue(std::make_pair(std::move(p), std::move(t)));
+    // Add captured stdout if necessary
+    faabric::util::SystemConfig& conf = faabric::util::getSystemConfig();
+    if (conf.captureStdout == "on") {
+        std::string moduleStdout = getCapturedStdout();
+        if (!moduleStdout.empty()) {
+            std::string newOutput = moduleStdout + "\n" + msg.outputdata();
+            msg.set_outputdata(newOutput);
 
-    // Check if we can add a thread
```
Shillaker, line 380: All this thread pooling stuff is handled in Faabric now

src/threads/ThreadState.cpp
```diff
@@ -53,45 +146,114 @@ int Level::getMaxThreadsAtNextLevel() const
...
-    int maxActiveLevels = std::max(msg.ompmal(), 1);
+    if (localThreadNum < 0) {
+        faabric::util::getLogger()->error(
+          "Local thread num negative: {} - {} @ {}",
+          msg->appindex(),
+          globalTidOffset,
+          depth);
+    }
+
+    assert(localThreadNum >= 0);
```
csegarragonz, line 242: Checking, throwing an error, and asserting seems a bit counter-intuitive/verbose to me. What do you think about using something like:
```cpp
#define assertm(exp, msg) assert(((void)msg, exp))
...
assertm(localThreadNum >= 0, fmt::format("Local thread num negative {} - {} @ {}", msg->appindex(), globalTidOffset, depth));
```
taken from [cppreference](https://en.cppreference.com/w/cpp/error/assert).

src/wavm/threads.cpp
```diff
@@ -134,25 +123,19 @@ WAVM_DEFINE_INTRINSIC_FUNCTION(env,
     const std::shared_ptr<spdlog::logger>& logger = faabric::util::getLogger();
     logger->debug("S - pthread_join - {} {}", pthreadPtr, resPtrPtr);
 
-    int returnValue;
+    // Await the chained thread
+    WAVMWasmModule* thisModule = getExecutingWAVMModule();
+    unsigned int callId = thisModule->chainedThreads[pthreadPtr];
+    logger->debug("Awaiting pthread: {} ({})", pthreadPtr, callId);
+    faabric::scheduler::Scheduler& sch = faabric::scheduler::getScheduler();
```
csegarragonz, line 130: I think we decided it was ok to use `auto&` for the scheduler's getter

tests/test/wasm/test_openmp.cpp
```diff
@@ -1,34 +1,32 @@
...
 
-    faabric::snapshot::SnapshotRegistry& reg =
-      faabric::snapshot::getSnapshotRegistry();
-    REQUIRE(reg.getSnapshotCount() == 0);
-
     // Make sure we have enough thread pool capacity
-    conf::FaasmConfig& conf = conf::getFaasmConfig();
-    int32_t defaultPoolSize = conf.moduleThreadPoolSize;
-    conf.moduleThreadPoolSize = 15;
+    faabric::util::SystemConfig& conf = faabric::util::getSystemConfig();
```
csegarragonz, line 22: Should we use `auto&` here as well? (Feel free to ignore)

src/wavm/WAVMWasmModule.cpp
```diff
@@ -851,29 +817,13 @@ uint32_t WAVMWasmModule::addFunctionToTable(Runtime::Object* exportedFunc)
...
-        if (boundUser != msg.user() || boundFunction != msg.function()) {
-            const std::string funcStr = faabric::util::funcToString(msg, true);
-            logger->error("Cannot execute {} on module bound to {}/{}",
-                          funcStr,
-                          boundUser,
-                          boundFunction);
-            throw std::runtime_error(
-              "Cannot execute function on module bound to another");
-        }
+        throw std::runtime_error("Module must be bound before executing");
```
csegarragonz, line 823: Maybe another use case for `assertm()`?

src/wasm/WasmModule.cpp
```diff
@@ -318,193 +362,64 @@ uint32_t WasmModule::getCurrentBrk()
...
+        // Vanilla function
+        returnValue = executeFunction(msg);
     }
 
-    // Wait
-    for (auto& t : openMPThreads) {
-        if (t.second.joinable()) {
-            t.second.join();
-        }
+    if (returnValue != 0) {
```
csegarragonz, line 405: `assertm()` maybe?

csegarragonz (approved):
LGTM. Made a small point on `SharedVars` and maybe introducing `assertm` a macro for assert + message

src/threads/ThreadState.cpp
```diff
@@ -1,22 +1,109 @@
...
 {}
 
+std::vector<uint32_t> Level::getSharedVars()
+{
+    return std::vector<uint32_t>(sharedVars, sharedVars + nSharedVars);
+}
+
+void Level::setSharedVars(uint32_t* ptr, int nVars)
+{
+    sharedVars = new uint32_t[nVars];
```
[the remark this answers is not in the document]
Shillaker: First off `sharedVars` is probably not a great name as they're really wasm offsets, so `sharedVarOffsets` might be better. Because they're wasm offsets, they'll always be `uint32_t`.

I'm not sure what sort of abstraction would make things any safer or clearer here. It would have to be serialisable so would need to be composed of primitives itself, so it could be a struct that held the value and size, e.g. 

```
struct SharedVarPtr {
    uint32_t val;
    size_t size = sizeof(uint32_t);
}
```

but I'm not sure that adds much clarity or extra safety (plus the `size` would always be constant). I could add a `#define SHARED_VAR_PTR_SIZE  sizeof(uint32_t)` at the top to make that part explicit, but it's only used in two places so I'm not sure it's worth it.

Any mistakes in the serialisation and deserialisation should be picked up in the tests, so I'm happy to leave this as is.

tests/test/wasm/test_openmp.cpp
Shillaker: We should use `auto` to reduce verbosity on very commonly used things, but not when it would impact the readability/ understanding of the code. The logger is a good example of where it's used in almost every file, its type name is long, and it's used in almost every file. Adding `auto` on the Scheduler is definitely ok in places where it's used a lot in the same file (e.g. tests), plus we only have one scheduler so you know what type it is. Given that we have two configs (faasm and faabric) I would say it's not ok to do it on `conf` as it might harm the readability, but it's probably not that big a deal either way.

src/threads/ThreadState.cpp
Shillaker: Hmm yes you're right, I'll just add an exception in this case.

src/wavm/WAVMWasmModule.cpp
Shillaker: Hmm it's a good question. I think `assertm` is probably never preferable to an `if` and an exception, other than the fact that it saves some lines. I do like the fact that `assertm` can be done in one line, hence doesn't clutter the code, but because it's removed in non-debug builds it risks missing errors in deployments. It would be nice to save some lines on checks like this though, so perhaps it should be replaced with an `assertm`-like macro that doesn't get removed.

I would say we should only use `assert` in places where we're checking things that, in theory, will never go wrong, e.g. checking arithmetic in a loop, testing that something isn't a nullptr in some internal accessor method. I guess I think of them as extensions of tests, where if you can look at it and say "ah that's obviously wrong". However, it's not clear-cut and perhaps we should create something like `assertm` but not actually using `assert`

PR #412, opened 2021-04-22
Bump faabric + Doc updates

csegarragonz:


.env
```diff
@@ -1,4 +1,5 @@
-FAASM_VERSION=0.5.15
-CPP_CLI_IMAGE=faasm/cpp-sysroot:0.0.24
+FAASM_VERSION=0.5.16
+FAASM_CLI_IMAGE=faasm/cli:0.5.16
```
csegarragonz, line 2: There was no default for `FAASM_CLI_IMAGE`

CMakeLists.txt
```diff
@@ -138,7 +138,7 @@ if (${FAASM_PERF_PROFILING})
     # Don't omit frame pointers (lets perf get better stack traces)
     set(CMAKE_CXX_FLAGS_DEBUG "${CMAKE_CXX_FLAGS_DEBUG} -fno-omit-frame-pointer")
 else ()
-    find_package(LLVM REQUIRED CONFIG)
+    find_package(LLVM 10 REQUIRED CONFIG)
```
csegarragonz, line 141: I needed this as I have various `LLVM` versions in my local machine. I think it's a good idea to pin versions.

.env
```diff
@@ -1,4 +1,5 @@
...
+CPP_CLI_IMAGE=faasm/cpp-sysroot:0.0.23
```
Shillaker, line 3: Why has this version gone down? `0.0.24` is the latest isn't it?

docs/development.md
```diff
@@ -266,6 +266,19 @@ This will mount your local build of Faasm inside a local cluster defined in
 Before doing so, make sure you've completely removed any existing faasm
 containers.
 
+Docker compose is well-suited for development as it allows to patch many aspects
+of your cluster without ever shutting down the deployment (which you may do
+by running `docker-compose down`).
```
Shillaker, line 271: Do we need this paragraph? Sounds like a general sales pitch for docker-compose

docs/development.md
```diff
@@ -266,6 +266,19 @@ This will mount your local build of Faasm inside a local cluster defined in
...
+
+### Scaling services
+
+To scale up or down a service you may run:
+```
+docker-compose up -d --scale <service>=<NUM_REPLICAS> [--force-recreate/no-recreate]
+```
```
Shillaker, line 278: This is just repeating the docker-compose documentation. The only service we might want to scale is the worker, so you could change this section to something like "Running with multiple workers"

docs/development.md
```diff
@@ -312,3 +325,15 @@ docker-compose restart worker
...
+Note that you can also update the environment in a service by modifying the
+`docker-compose.yml` file, and re-running:
+```
+docker-compose up -d
+```
+
+To check that the env. variable has been updated, run:
+```
+docker-compose exec <service> env
+```
```
Shillaker, line 339: Again this is just a repeat of how to use docker-compose. We could hyperlink the docker-compose docs somewhere near the top if you think it's useful, otherwise I think we can assume people either already know this, or can google it.

Shillaker (changes requested):
Version bump looks good, the doc updates might be a bit excessive. Remember that we don't want to repeat documentation for common tools like docker-compose. To keep things concise and avoid the risk of rotting, we should only include stuff that's specific to Faasm (and assume people know how to use popular tools or can search for their docs). If we see anywhere the docs are getting a bit long-winded in this regard, we should cut them down.

.env
csegarragonz: My mistake, thanks for spotting this.

Shillaker (approved):

PR #415, opened 2021-05-13
Move in Faasm-specific config from Faabric

Shillaker:
Updates to fit with [Faabric PR](https://github.com/faasm/faabric/pull/89).

- Moved config into `FaasmConfig` object
- Added `function_utils.cpp` to hold what was taken from Faabric's function utils
- Changed use of Faabric config to Faasm config where necessary
- Removed use of `auto` on configs to make it clear which one is being used

### DIFF
--- a/include/conf/function_utils.h
+++ b/include/conf/function_utils.h
@@ -1,0 +1,64 @@
+#pragma once
+
+#include <faabric/proto/faabric.pb.h>
+#include <faabric/util/exception.h>
+#include <string>
+#include <vector>
+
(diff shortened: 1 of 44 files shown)


include/conf/function_utils.h
```diff
@@ -0,0 +1,64 @@
+#pragma once
```
Shillaker, line 1: This is cleaved from Faabric's `func.cpp` util.

csegarragonz (approved):

PR #416, opened 2021-05-14
Distributed tests

Shillaker:
Similar to the distributed tests used in Faabric [introduced in this PR](https://github.com/faasm/faabric/pull/88) implementations will start out looking similar but inevitably diverge.

We can merge this to get things going, then I'll add some more tests in another PR.

### DIFF
--- a/dist-test/run.sh
+++ b/dist-test/run.sh
@@ -1,0 +1,40 @@
+#!/bin/bash
+
+set -e
+
+export PROJ_ROOT=$(dirname $(dirname $(readlink -f $0)))
+pushd ${PROJ_ROOT}/dist-test >> /dev/null
+
+# Set up image name
+if [[ -z "${FAASM_CLI_IMAGE}" ]]; then
+    VERSION=$(cat ../VERSION)
+    export FAASM_CLI_IMAGE=faasm/cli:${VERSION}
+fi
+
+if [ "$1" == "local" ]; then
+    INNER_SHELL=${SHELL:-"/bin/bash"}
+
+    # Start everything in the background
+    docker-compose \
+        up \
+        --no-recreate \
+        -d \
+        master
+
+    # Run the CLI
+    docker-compose \
+        exec \
+        master \
+        ${INNER_SHELL}
+else
+    # Run the tests directly
+    docker-compose \
+        run \
+        master \
+        /build/faasm/bin/dist_tests
+
+    docker-compose \
+        stop
+fi
+
+popd >> /dev/null
--- a/tests/dist/CMakeLists.txt
+++ b/tests/dist/CMakeLists.txt
@@ -1,16 +1,18 @@
 file(GLOB_RECURSE TEST_FILES ${CMAKE_CURRENT_LIST_DIR} test_*.cpp)
 
-include_directories(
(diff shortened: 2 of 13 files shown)


dist-test/run.sh
```diff
@@ -0,0 +1,40 @@
...
+
+if [ "$1" == "local" ]; then
+    INNER_SHELL=${SHELL:-"/bin/bash"}
+
+    # Start everything in the background
+    docker-compose \
+        up \
+        --no-recreate \
+        -d \
+        master
```
csegarragonz, line 22: Do you think we could also add a parameter to set the number of workers? It may be useful to have an arbitrary number of workers.

csegarragonz, line 22: Do you think we could also add a parameter to set the number of workers? It may be useful to have an arbitrary number of workers.

tests/dist/CMakeLists.txt
```diff
@@ -0,0 +1,27 @@
+file(GLOB_RECURSE TEST_FILES ${CMAKE_CURRENT_LIST_DIR} test_*.cpp)
+
+include_directories(
```
csegarragonz, line 3: Could we use:
```
target_include_directories(dist_tests ...)
```

tests/dist/CMakeLists.txt
```diff
@@ -0,0 +1,27 @@
+file(GLOB_RECURSE TEST_FILES ${CMAKE_CURRENT_LIST_DIR} test_*.cpp)
```
csegarragonz, line 1: Could we avoid the globbing here? I know it's in our To-Do list, but we might as well enforce it already.

tests/dist/fixtures.h
```diff
@@ -0,0 +1,46 @@
+#pragma once
```
csegarragonz, line 1: :+1:

csegarragonz (approved):
LGTM. And I don't remember the PR you link to, but you must have had a good time doing it :wink:

dist-test/run.sh
Shillaker: Yes perhaps in future, I'm not there's much benefit though, what specifically do you think it would achieve? If we make it dynamic, the tests would also need to access that parameter and adjust to it when running checks, which makes it harder to write them. At the moment just having one other host with a hard-coded API makes it a lot simpler (and covers a lot of the stuff we'd like to test)

tests/dist/CMakeLists.txt
Shillaker: I agree with removing globbing in the main source, but in the tests we won't know if we've missed a file (e.g. if I add a new test to the source tree and forget to put it in the `CMakeLists.txt` it won't run), so I actually would argue for keeping the globbing in the tests. The only downside to doing it is that we sometimes have to rerun `cmake` to pick up new tests right? I personally don't find that too much of a hassle given the benefits.

dist-test/run.sh
csegarragonz: :+1:

tests/dist/CMakeLists.txt
csegarragonz: :+1:

PR #418, opened 2021-05-20
Extend distributed tests to cover chaining and threading

Shillaker:
This builds out the distributed testing to include invoking functions, chaining and threading. It also hooks Faaslets into Faabric's dirty page checking (via the `snapshot()` method). The OpenMP function invoked in the tests doesn't actually create any dirty pages, so it's not checking the correctness, just adding the scaffolding (I'll put a "proper" test in the next PR to keep this one small).

Changes:

- Add running of wasm functions to distributed tests.
- Add a couple of tests for simple function execution and chaining.
- Add test for distributed multi-threading.
- Add a Faaslet `snapshot()` method to hook into the Faabric dirty page checking.
- Bind Faaslets in the constructor rather than just before execution time. This is to allow Faaslets to be used straight after construction (necessary to perform snapshot/ restore tasks before execution).


dist-test/build.sh
```diff
@@ -14,6 +14,7 @@ fi
 # Run the build
 docker-compose \
     run \
+    --rm \
```
Shillaker, line 17: Without this a new container is created on every run. Annoying when iterating locally.

.github/workflows/tests.yml
```diff
@@ -171,16 +171,22 @@ jobs:
...
         run: git submodule update --init faabric
-      # --- Build and test ---
+      # --- Set up and run ---
       - name: "Build distributed tests"
         run: ./dist-test/build.sh
+      - name: "Download CPP wasm"
+        uses: actions/download-artifact@v2
+        with:
+          name: cpp-wasm
+          path: ./dist-test/build/faasm-local/wasm
```
Shillaker, line 189: This downloads the compiled wasm that we use in the normal tests into the distributed test container.

Shillaker:
Dividing into several smaller PRs to avoid changing too much at once.

PR #420, opened 2021-05-22
Bump faabric + Changes for ZeroMQ

csegarragonz:
In this PR I bump `faabric`'s version and add a couple of changes needed to integrate with ZeroMQ as introduced in faasm/faabric#85.

Most notably, there was a race condition in the MPI world shutdown that has been fixed upstream in faasm/faabric#98 and faasm/faabric#100. It was again one of this race conditions that only apppear when the system is slow enough (i.e. GHA).

Tests are passing so I feel this is ready for review, I'm handling the image rebuild in the background.

After this is merged in I'll create and publish a release for Faasm 0.6.0.


csegarragonz:
Release build failing with:
```
fatal: write error: No space left on device
```
I have manually built and pushed the images to docker hub.

faasmcli/requirements.txt
```diff
@@ -1,5 +1,6 @@
 ansible
 awscli
+black
```
csegarragonz, line 3: I just realized that we installed black as part of `grpc-root` [here](https://github.com/faasm/faabric/blob/0b863976bc8b88ed904cc1ae89dce9a63a514685/docker/grpc-root.dockerfile#L22).

Shillaker (approved):

PR #429, opened 2021-06-08
Fix race conditions in caching of wasm modules

Shillaker:
Fixes a race condition on map accesses in WAVM-related caching.

### DIFF
--- a/src/wavm/WAVMModuleCache.cpp
+++ b/src/wavm/WAVMModuleCache.cpp
@@ -29,29 +29,29 @@ wasm::WAVMWasmModule& WAVMModuleCache::getCachedModule(
   const faabric::Message& msg)
 {
     std::string key = faabric::util::funcToString(msg, false);
-    if (cachedModuleMap.count(key) == 0) {
+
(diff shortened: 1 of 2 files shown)


src/wavm/WAVMModuleCache.cpp
```diff
@@ -29,29 +29,29 @@ wasm::WAVMWasmModule& WAVMModuleCache::getCachedModule(
...
-
-    // Instantiate the base module
-    wasm::WAVMWasmModule& module = cachedModuleMap[key];
-    module.bindToFunction(msg, false);
+    {
+        // Note that we need a shared lock here to avoid a race condition on
+        // initialising the module
+        faabric::util::SharedLock lock(mx);
+        return cachedModuleMap[key];
+    }
```
Shillaker, line 54: Note that I inlined the `initialiseCachedModule` function here to simplify things so the diff looks a little longer than it should.

csegarragonz (approved):
I am running into an assertion failure very often now:
```
other.compartment != nullptr
```
hopefuly this fixes it :D

PR #432, opened 2021-06-14
Use S3 API and Minio for file storage

Shillaker:
We currently have a very verbose internal interface that supports multiple file storage backends (e.g. different cloud object stores). However, it's only ever used with our dummy HTTP fileserver, which can't handle larger files. 

Given that the S3 API is supported by most cloud object storage, we can switch to supporting only the S3 API for storage and use [minio](https://min.io/) locally and in K8s deployments.

We used to have an S3 storage client which got removed in https://github.com/faasm/faasm/pull/227, and can be resurrected. 

This PR just covers resurrecting the old S3 stuff and getting it to work with minio locally. Other PRs will port and test the storage interface.

Changes:

- Add the [`aws-cpp-sdk`](https://github.com/aws/aws-sdk-cpp) as an external CMake project (note that their S3 client can be used with any S3-compatible storage including minio).
- Resurrect the `S3Wrapper` class and associated tests (and refactor to reduce repetition).
- Add the credentials and bucket name to the `FaasmConfig`
- Add a `minio` container where necessary (when switching fully to S3 I'll add the associated k8s config)

Unfortunately because GHA apparently doesn't support custom commands for `service` containers, I've had to add our own `faasm/minio` container for now.


docker/worker.dockerfile
```diff
@@ -1,8 +1,6 @@
 ARG FAASM_VERSION
 FROM faasm/base:${FAASM_VERSION}
 
-COPY . /usr/local/code/faasm
-
```
Shillaker, line 5: We stopped doing `COPY .` a while ago as it can pull in changes specific to the local machine, and bloat the size of the image. The code is set up in the `base` image by checking out the relevant tag, so this is also not needed.

faasmcli/faasmcli/tasks/docker_tasks.py
```diff
@@ -66,8 +66,6 @@ def build(ctx, c, nocache=False, push=False):
     faasm_ver = get_faasm_version()
 
     for container in c:
-        # Check if we need to template a special dockerignore file It seems the
-        # dockerignore file needs to be at <dockerfile_path>.dockerignore
```
Shillaker, line 70: This comment is no longer valid and should have been removed in the PR that removed the custom dockerignore feature.

include/storage/S3FileLoader.h
```diff
@@ -0,0 +1,133 @@
+#pragma once
+
+#include <conf/FaasmConfig.h>
+#include <faabric/util/config.h>
+#include <faabric/util/func.h>
+#include <storage/FileLoader.h>
+#include <storage/S3Wrapper.h>
+
+namespace storage {
+class S3FileLoader final : public FileLoader
```
Shillaker, line 10: The very verbose `FileLoader` interface is something I'll simplify in a later PR.

src/storage/S3Wrapper.cpp
```diff
@@ -0,0 +1,227 @@
...
+    SPDLOG_DEBUG("Listing buckets");
+    auto response = client.ListBuckets();
+    CHECK_ERRORS(response);
+
+    Aws::Vector<Bucket> bucketObjects = response.GetResult().GetBuckets();
+
+    std::vector<std::string> bucketNames;
+    for (auto const& bucketObject : bucketObjects) {
+        const Aws::String& awsStr = bucketObject.GetName();
+        bucketNames.emplace_back(awsStr.c_str());
```
csegarragonz, line 126: Very minor, but given that `bucketNames` is a vector of `std::string`s and `emplace_back` constructs them in place, the following would feel more natural to me:
```cpp
bucketNames.emplace_back(awsStr.c_str(), awsStr.size());
```

src/storage/S3Wrapper.cpp
```diff
@@ -0,0 +1,227 @@
...
+    auto response = client.ListObjects(request);
+
+    CHECK_ERRORS(response);
+
+    Aws::Vector<Object> keyObjects = response.GetResult().GetContents();
+
+    std::vector<std::string> keys;
+    for (auto const& keyObject : keyObjects) {
+        const Aws::String& awsStr = keyObject.GetKey();
+        keys.emplace_back(awsStr.c_str());
```
csegarragonz, line 145: Same comment as before, but feel free to ignore.

csegarragonz (approved):

src/storage/S3Wrapper.cpp
Shillaker: Yeah I guess it's equivalent provided the strings are properly null-terminated (albeit with a miniscule performance difference required to locate that null terminator). In this case they're coming from the `Aws::String` class which will ensure proper null termination. I'll switch in the next PR.

PR #435, opened 2021-06-15
Faabric update: per-host flushing, dirty pages integration

Shillaker:
Change to `FaasletFactory::flushHost` from `Faaslet::flush` for host-wise flushing: https://github.com/faasm/faabric/pull/112

Add Faaslet `snapshot()` method to hook into Faabric dirty page checking: https://github.com/faasm/faabric/pull/96


Shillaker:
Merged into https://github.com/faasm/faasm/pull/434

PR #439, opened 2021-06-17
Add distributed test for chaining

Shillaker:


dist-test/docker-compose.yml
```diff
@@ -52,7 +52,7 @@ services:
       - ${FAASM_LOCAL_DIR}:/usr/local/faasm
     working_dir: /build/faasm
     environment:
-      - LOG_LEVEL=info
+      - LOG_LEVEL=debug
```
csegarragonz, line 55: Sorry for necroposting, just updating on last week's changes. Are we sure we want this set to `debug`?

Shillaker: Yes, for now given that there are only three of them and they fail so often we can stick with debug

PR #440, opened 2021-06-17
Add distributed test for pthreads and OpenMP

Shillaker:
Continuing to extend the distributed testing with checks for pthreads and OpenMP. 

Both tests run multi-threaded functions that will have one or more threads exectuing on the main host, and one or more on the remote host. Threads on all hosts will then write to shared memory, and the main thread will check that these writes are merged across hosts before the function completes.

Changes:

- Added new distributed tests.
- Added missing `contextdata` and `subtype` to messages when offloading threads to other hosts.
- Fixed bug with serialisation and deserialisation of OpenMP levels (the (de)serialisation of the array of `sharedVarOffsets` just didn't work before). Now just copies the underlying bytes of the class instance rather than initialising individual fields.
- Added more detailed logging around OpenMP level (de)serialisation and snapshot memory growth/ shrinking.


dist-test/build.sh
```diff
@@ -14,6 +14,7 @@ fi
 # Run the build
 docker-compose \
     run \
+    --rm \
```
Shillaker, line 17: Avoids leaving a load of orphaned containers when repeatedly running this.

src/wavm/WAVMModuleCache.cpp
```diff
@@ -37,8 +37,7 @@ wasm::WAVMWasmModule& WAVMModuleCache::getCachedModule(faabric::Message& msg)
         // Re-check condition
         if (cachedModuleMap.find(key) == cachedModuleMap.end()) {
 
-            SPDLOG_DEBUG(
-              "WAVM module cache initialising {} (tid {})", key, gettid());
+            SPDLOG_DEBUG("WAVM module cache initialising {}", key);
```
Shillaker, line 40: The tid is now included in all logging statements.

src/wavm/openmp.cpp
```diff
@@ -408,9 +408,9 @@ WAVM_DEFINE_INTRINSIC_FUNCTION(env,
     // Prepare arguments for main thread and all others
     std::vector<IR::UntaggedValue> mainArguments = { 0, argc };
     if (argc > 0) {
-        // Build list of pointers to shared variables
-        U32* sharedVarsPtr =
-          Runtime::memoryArrayPtr<U32>(memoryPtr, argsPtr, argc);
+        // Build list of offsets to shared variables
+        uint32_t* sharedVarsPtr =
+          Runtime::memoryArrayPtr<uint32_t>(memoryPtr, argsPtr, argc);
```
Shillaker, line 413: Trying to gradually move away from using WAVM custom types (e.g. `U32`)

tests/utils/worker_utils.cpp
```diff
@@ -4,7 +4,7 @@
 #include <faabric/proto/faabric.pb.h>
 #include <faabric/runner/FaabricMain.h>
 #include <faabric/scheduler/FunctionCallClient.h>
-#include <faabric/scheduler/SnapshotClient.h>
+#include <faabric/snapshot/SnapshotClient.h>
```
Shillaker, line 7: This was moved around in the Faabric update.

include/threads/ThreadState.h
```diff
@@ -21,31 +21,34 @@ class Level
     uint32_t id;
 
     // Number of nested OpenMP constructs
-    int depth = 0;
+    int32_t depth = 0;
```
Shillaker, line 24: Because the raw bytes underlying this object are serialised, sent across the wire and deserialised, I wanted to be explicit about the width of each field.

include/threads/ThreadState.h
```diff
@@ -21,31 +21,34 @@ class Level
...
 
     // Offset for the global thread numbers at this level
     int32_t globalTidOffset = 0;
 
     uint32_t nSharedVarOffsets = 0;
     uint32_t* sharedVarOffsets;
 
-    Level(int numThreadsIn);
+    static std::shared_ptr<Level> deserialise(
+      const std::vector<uint8_t>& bytes);
```
Shillaker, line 49: Previously `deserialise` was an instance method on the `Level` class (i.e. create the `Level`, _then_ deserialise the contents from some bytes), but it makes more sense to have it as a static factory-like method.

src/threads/ThreadState.cpp
```diff
@@ -178,28 +185,41 @@ std::vector<uint8_t> Level::serialise()
...
-    pushedThreads = other->pushedThreads;
-    wantedThreads = other->wantedThreads;
-
-    globalTidOffset = other->globalTidOffset;
-
-    nSharedVarOffsets = other->nSharedVarOffsets;
-    sharedVarOffsets = new uint32_t[nSharedVarOffsets];
-    std::memcpy(sharedVarOffsets,
-                other->sharedVarOffsets,
-                nSharedVarOffsets * sizeof(uint32_t));
```
Shillaker, line 202: Relying on directly moving the underlying bytes around avoids having to explicitly set all the fields of the object.

src/threads/ThreadState.cpp
```diff
@@ -178,28 +185,41 @@ std::vector<uint8_t> Level::serialise()
...
+    // Copy the top section of bytes into a new instance (relying on the
+    // C++ class memory layout placing members at the top)
```
Shillaker, line 208: This assumption is key to the serialisation and deserialisation, but it's a safe one to make AFAICT (i.e. a pointer to a class instance is like a struct with the same fields laid out end-to-end).

src/threads/ThreadState.cpp
```diff
@@ -11,6 +13,8 @@
 
 using namespace faabric::util;
 
+#define LEVEL_WAIT_TIMEOUT_MS 20000
```
Shillaker, line 16: This is somewhat arbitrary, it will be hit if a thread dies and doesn't hit the wait section, but it is preferable to have a clear error after a reasonable wait rather than an infinite hang.

src/threads/ThreadState.cpp
```diff
@@ -157,8 +162,15 @@ void Level::masterWait(int threadNum)
...
-            nowaitCv->wait(lock);
+        auto timePoint = std::chrono::system_clock::now() +
+                         std::chrono::milliseconds(LEVEL_WAIT_TIMEOUT_MS);
+
+        if (!nowaitCv->wait_until(lock, timePoint, [&] {
+                return nowaitCount->load() >= numThreads - 1;
+            })) {
+
+            SPDLOG_ERROR("Level {} master wait timed out", id);
+            throw std::runtime_error("Level wait on master timed out");
```
Shillaker, line 173: This is adding a timeout on the wait rather than having an infinite wait.

src/threads/ThreadState.cpp
```diff
@@ -278,4 +311,17 @@ int Level::getGlobalThreadNum(faabric::Message* msg)
...
+    std::stringstream ss;
+    ss << "Level " << id << " depth=" << depth << " threads=" << numThreads
+       << " shared={ ";
+    for (int i = 0; i < nSharedVarOffsets; i++) {
+        ss << sharedVarOffsets[i] << " ";
+    }
+    ss << "}";
+
+    return ss.str();
+}
```
Shillaker, line 326: This function is only used for trace logging purposes, so we don't need to worry about performance or efficiency.

src/wavm/threads.cpp
```diff
@@ -21,10 +21,6 @@ using namespace WAVM;
 
 namespace wasm {
 
-// Record of whether this thread has already got a snapshot being used to spawn
-// other threads.
-static thread_local std::string currentSnapshotKey;
```
Shillaker, line 26: This isn't used anywhere.

src/threads/ThreadState.cpp
```diff
@@ -178,28 +196,43 @@ std::vector<uint8_t> Level::serialise()
     size_t thisSize = sizeof(Level);
     thisSize += nSharedVarOffsets * sizeof(uint32_t);
 
+    std::vector<uint8_t> bytes(thisSize, 0);
+
+    // Copy the level without the shared offsets (relying on the C++ class
+    // memory layout placing members at the top
```
csegarragonz, line 202: small typo. missing the closing parenthesis.

tests/dist/chaining/test_openmp.cpp
```diff
@@ -0,0 +1,45 @@
+#include <catch2/catch.hpp>
```
csegarragonz, line 1: Slightly unrelated point: why do this tests live in `tests/dist/chaining`? The name `chaining` seems a bit odd to me.

csegarragonz (approved):

PR #441, opened 2021-06-17
Add section logging to test loggers

Shillaker:
Switch to standardised test logging based on https://github.com/faasm/faabric/pull/116


csegarragonz:
Talking about duplication, I have noticed that, even when there is no actual `SECTION` keyword, there's one implicit section, so we have all no-section tests with a duplicated log. Are we ok with this?

Shillaker:
Ooh very good spot, I hadn't noticed that. I'll see if I can avoid it if there's only one section.

csegarragonz (approved):

PR #443, opened 2021-06-24
Fix SGX build

csegarragonz:
This PR is the first of a set of PRs that aim to fix the current SGX implementation in master. As it stands now:
1. Compilation is disabled as it fails
2. The only test is commented out

This PR fixes 1.

I also add some comments in the header files, and do some clean up (happy to revert, as not directly related to fixing the build, and somewhat arbitrary).

CC-ing @dgoltzsche as well


cmake/ExternalProjects.cmake
```diff
@@ -15,8 +15,8 @@ FetchContent_Declare(wavm_ext
 )
 
 FetchContent_Declare(wamr_ext
-    GIT_REPOSITORY "https://github.com/J-Heinemann/wasm-micro-runtime"
-    GIT_TAG "5a3db31c096ec4d8a7b8b639aeaecd3fdda06b30"
+    GIT_REPOSITORY "https://github.com/faasm/wasm-micro-runtime"
```
csegarragonz, line 18: Both references point to the same commit.

src/sgx/CMakeLists.txt
```diff
@@ -42,41 +41,41 @@ set(THREADS_PREFER_PTHREAD_FLAG ON)
...
-        ${IWASM_COMMON_SOURCE}
-        ${IWASM_INTERP_SOURCE}
-        ${LIBC_BUILTIN_SOURCE}
-        ${LIBC_WASI_SOURCE}
-        ${MEM_ALLOC_SHARED_SOURCE}
-        ${PLATFORM_SHARED_SOURCE}
-        ${UTILS_SHARED_SOURCE}
-        )
+# Let WAMR do the including and importing of the sources
+include (${WAMR_ROOT_DIR}/build-scripts/runtime_lib.cmake)
```
csegarragonz, line 77: This seems to be how it's done in [WAMR](https://github.com/bytecodealliance/wasm-micro-runtime/blob/main/product-mini/platforms/linux-sgx/CMakeLists.txt#L86)'s example.

This is, (1) set variables, (2) include just one `*.cmake` file.

src/runner/CMakeLists.txt
```diff
@@ -11,12 +11,18 @@ target_link_libraries(func_runner ${RUNNER_LIBS})
...
-
-add_executable(codegen_func codegen_func.cpp)
-target_link_libraries(codegen_func ${RUNNER_LIBS})
-
 # Main entrypoint for worker nodes
 add_executable(pool_runner pool_runner.cpp)
 target_link_libraries(pool_runner ${RUNNER_LIBS})
+
+set(CODEGEN_LIBS
+        wavmmodule
```
csegarragonz, line 19: `faaslet_lib` in `RUNNER_LIBS` would make the compilation of the codegen targets fail

include/sgx/enclaveConfig.h
```diff
@@ -0,0 +1,17 @@
...
+
+/* Main hardcoded memory sizes */
+#define ONE_KB_BYTES 1024
+#define ONE_MB_BYTES (1024 * 1024)
+#define FAASM_SGX_WAMR_HEAP_SIZE (32 * ONE_MB_BYTES)
+#define FAASM_SGX_WAMR_MODULE_ERROR_BUFFER_SIZE 128
+#define FAASM_SGX_WAMR_INSTANCE_DEFAULT_HEAP_SIZE (8 * ONE_KB_BYTES)
+#define FAASM_SGX_WAMR_INSTANCE_DEFAULT_STACK_SIZE (8 * ONE_KB_BYTES)
+
+/* Thread Control Structure (TCS)
```
Shillaker, line 11: This comment seems a bit long for a single `#define` and includes some general info about SGX, perhaps a shorter comment saying specifically what this parameter does in the context of our application, then a link to the TCS docs would be better.

include/sgx/faasm_sgx_native_symbols_wrapper.h
```diff
@@ -1,14 +1,18 @@
...
 #include <iwasm/include/wasm_export.h>
 #include <sgx.h>
 #include <sgx_defs.h>
 
 // Length of used native symbols
-#define FAASM_SGX_NATIVE_SYMBOLS_LEN 26
+#define FAASM_SGX_NATIVE_SYMBOLS_LEN 25
 #define FAASM_SGX_WASI_SYMBOLS_LEN 6
 
+/* Native symbols definition
```
Shillaker, line 11: Re. offline comments I think this is covered by the name of the file.

src/sgx/SGXWAMRWasmModule.cpp
```diff
@@ -86,6 +87,11 @@ void SGXWAMRWasmModule::doBindToFunction(faabric::Message& msg, bool cache)
                      faasmSgxErrorString(returnValue));
         throw std::runtime_error("Unable to load WASM module");
     }
+
+    // Set up the thread stacks
+    // 28/06/2021 - Threading is not supported in SGX-WAMR. Thus, we don't care
+    // what we put in this vector.
+    threadStacks.push_back(-1);
```
Shillaker, line 94: Can it be left empty or is there a check that it's the right size? Might be good to say why we have to put a value in there if we don't care about it.

src/sgx/faasm_sgx_enclave.cpp
```diff
@@ -23,6 +24,13 @@
 #define WASM_CTORS_FUNC_NAME "__wasm_call_ctors"
 #define WASM_ENTRY_FUNC "_start"
 
+/* Enclave code for a WAMR enclave running in Faasm
```
Shillaker, line 27: Again, this should be covered by the file name.

include/sgx/enclaveConfig.h
```diff
@@ -0,0 +1,17 @@
+#pragma once
+
+/* Main hardcoded memory sizes */
```
Shillaker, line 3: Single-line comments should be `//`.

src/sgx/faasm_sgx_enclave.cpp
```diff
@@ -336,6 +338,12 @@ extern "C"
         return FAASM_SGX_SUCCESS;
     }
 
+    /* Load WebAssembly module to the enclave
```
Shillaker, line 341: This comment title should be covered by the name of the function.

src/sgx/faasm_sgx_enclave.cpp
```diff
@@ -417,30 +421,30 @@ extern "C"
...
-            ocall_printf(wamr_error_buffer);
-#endif
+            SGX_DEBUG_LOG(wamr_error_buffer);
             return FAASM_SGX_WAMR_MODULE_INSTANTIATION_FAILED;
         }
 
         return FAASM_SGX_SUCCESS;
     }
 
+    /* Initialise WebAssembly Micro Runtime inside the enclave.
```
Shillaker, line 431: Again, repeating the function name.

src/sgx/faasm_sgx_enclave.cpp
```diff
@@ -417,30 +421,30 @@ extern "C"
...
-#if (FAASM_SGX_DEBUG)
-        // Set our ocall_printf as default printf function if FAASM_SGX_DEBUG ==
-        // 1
-        os_set_print_function((os_print_function_t)ocall_printf);
-#endif
+        os_set_print_function((os_print_function_t)SGX_DEBUG_LOG);
 
         // Initialize FAASM-SGX TCS
+        // Note - currently we just support single-threaded execution, thus
+        // only one TCS slot.
```
Shillaker, line 443: If the value of `FAASM_SGX_INIT_TCS_SLOTS` was increased, would this comment become invalid? I feel like the one in `enclaveConfig.h` is probably enough.

cmake/ExternalProjects.cmake
Shillaker: I thought that we stopped using branch names because CMake wouldn't update properly? Should the `faasm` here not be the commit hash? If the problem with branch names is fixed, can we stop using commit hashes elsewhere?

include/sgx/SGXWAMRWasmModule.h
```diff
@@ -48,6 +57,13 @@ extern "C"
 }
 
 namespace wasm {
+/* Interface to interact with a WAMR runtime running inside an enclave.
```
Shillaker, line 60: The title of this comment should be captured in the class name.

src/sgx/CMakeLists.txt
Shillaker: I think including that link in the source could be useful for reference in future.

src/sgx/CMakeLists.txt
```diff
@@ -42,41 +41,41 @@ set(THREADS_PREFER_PTHREAD_FLAG ON)
...
+# --------------------------------------------------------
+
+# Set target platform details
+set (WAMR_BUILD_PLATFORM "linux-sgx")
+set (WAMR_BUILD_TARGET X86_64)
 
-if(FAASM_SGX_WAMR_AOT_MODE)
+# Set WAMR include flags
+# 28/06/2021 - (fast) interpreter, AOT, libc, wasi, and pthread are enabled by
+# default in WAMR's SGX port.
```
Shillaker, line 54: Is there a link to where this happens in the WAMR cmake files? Could we include a link? Although it's got a date on it, this comment is likely to become outdated and having a link to where it's actually done would be more time-proof.

Shillaker (changes requested):
Nice, great work resurrecting this. Some tangential comments on comments.

src/sgx/SGXWAMRWasmModule.cpp
csegarragonz: Yes, I also hesitated on this one. There's an assertion [here](https://github.com/faasm/faasm/blob/fe5cbb30c7add9d139f5fa7c84dc2e0fa0ed7d7b/src/wasm/WasmModule.cpp#L341) that makes us have to add at least one element.

I feel the assertion must stay, as this thread management is something we'd desirably have in all our runtimes. Thus it makes more sense to deliberately override it for SGX. In an ideal scenario, we'd support threading in SGX as well.

src/sgx/CMakeLists.txt
```diff
@@ -11,10 +11,7 @@ set(CMAKE_ASM_NASM_FLAGS -felf64)
 set(CMAKE_ASM_NASM_COMPILE_OBJECT "<CMAKE_ASM_NASM_COMPILER> <INCLUDES> ${CMAKE_ASM_NASM_FLAGS} -o <OBJECT> <SOURCE>")
 
 # FAASM-SGX enclave configuration
-set(FAASM_SGX_INIT_TCS_SLOTS 2)
 set(FAASM_SGX_ENCLAVE_PATH "${CMAKE_RUNTIME_OUTPUT_DIRECTORY}/faasm_sgx_enclave.sign.so")
-set(FAASM_SGX_WAMR_BUILDIN_LIBC 1)
```
csegarragonz, line 16: I have decided these options are too low-level and, most probably, will break the code if changed. Thus I delete them.

cmake/ExternalProjects.cmake
csegarragonz: You are right. The problem with pining to branch names is that CMake has no way of telling if it has to update the dependency or not (i.e. there are newer commits in the branch). Through commit hashes it does.

Shillaker (approved):
Awesome, LGTM 👍

src/sgx/SGXWAMRWasmModule.cpp
dgoltzsche: What exactly does us stop from supporting multithreading in WAMR in SGX? Does WAMR-SGX itself lack support?

PR #444, opened 2021-06-28
Fix SGX test

csegarragonz:
This PR fixes the existing SGX test (i.e. uncomments it), and ensures it passes in interpreter mode.

The test also passes in AoT mode, but I can't decide how we should test both.


CMakeLists.txt
```diff
@@ -9,14 +9,14 @@ option(FAASM_SELF_TRACING "Turn on system tracing using the logger" OFF)
 option(FAASM_PERF_PROFILING "Turn on profiling features as described in debugging.md" OFF)
 
 # WAMR and interpreter
-option(WAMR_INTERPRETER_MODE "Turns interpreter on / off (vs. AoT)" OFF)
+option(FAASM_WAMR_AOT_MODE "Turns on AoT mode in WAMR" ON)
```
csegarragonz, line 12: Now the behaviour for SGX and non-SGX WAMR in terms of options is the same:
* Interpreter mode is always built.
* If opt-in, AoT support is also built.
* If both available, will run in AoT mode.

src/sgx/CMakeLists.txt
```diff
@@ -53,8 +53,15 @@ set (WAMR_BUILD_TARGET X86_64)
 # Set WAMR include flags
 # (fast) interpreter, AOT, libc, wasi, and pthread are enabled by default
 
+# Enabling the spec tests performs stricter checks. E.g. it will error if can't
+# link an imported function
+if (FAASM_SGX_DEBUG)
+    add_definitions (-DWASM_ENABLE_SPEC_TEST)
```
csegarragonz, line 59: Adding this definition makes WAMR error out if it fails to link an imported function.

CMakeLists.txt
```diff
@@ -9,14 +9,15 @@ option(FAASM_SELF_TRACING "Turn on system tracing using the logger" OFF)
...
 
 # SGX functionality
 # These options customise the SGX features _provided_ SGX is found
 option(FAASM_SGX_SIM_MODE "Turn on SGX sim mode" ON)
 option(FAASM_SGX_ATTESTATION "Turn on attestation" OFF)
 option(FAASM_SGX_XRA "Turn on support for XRA" OFF)
-option(FAASM_SGX_WAMR_AOT_MODE "Turn on AoT mode in SGX" ON)
+option(FAASM_SGX_WAMR_AOT_MODE "Turn on AoT mode in SGX" OFF)
+set (FAASM_SGX_WAMR_AOT_MODE OFF)
```
csegarragonz, line 20: Until we decide how we choose between test modes, not having this `set` means that you must manually change the option, which is cached in `CMakeCache.txt`. I.e. re-build from scratch.

tests/utils/worker_utils.cpp
```diff
@@ -155,6 +155,10 @@ void execFuncWithPool(faabric::Message& call, bool clean, int timeout)
     faabric::Message result = sch.getFunctionResult(call.id(), 30000);
     REQUIRE(result.returnvalue() == 0);
 
+    if (result.outputdata().size() > 0) {
```
csegarragonz, line 158: I know there's a `execFuncWithStringResult`, but it felt unnatural to use it. And it also felt unnatural to pass an optional parameter to print the `outputdata`. I think printing it always as long as its got something in it (`results.outputdata()`, I mean), is acceptable.

Shillaker: The output isn't always human-redable, so this may pollute the logs with stuff that's hard to decipher. 

If you've found it useful though it's probably worth keeping, but:

- Make it at least debug level if not trace
- Prefix it with `Function <function_id> output: ` so that it doesn't look like junk

src/sgx/CMakeLists.txt
Shillaker: I think we should have this on all the time (both SGX and non-SGX WAMR). The WAVM implementation has this strict behaviour.

CMakeLists.txt
Shillaker: As discussed offline, I think we should remove interpreter completely.

include/sgx/faasm_sgx_native_symbols_wrapper.h
```diff
@@ -5,7 +5,7 @@
 #include <sgx_defs.h>
 
 // Length of used native symbols
-#define FAASM_SGX_NATIVE_SYMBOLS_LEN 25
+#define FAASM_SGX_NATIVE_SYMBOLS_LEN 33
```
Shillaker, line 8: This is so gross and is bound to be forgotten by the next person who comes to edit this code.

- What actually needs this? Is it something to do with SGX configuration?
- What error comes out if it's wrong? Can we make it more informative?
- When is the error thrown? Can we do it at initialisation time rather than waiting till it's more deeply nested?

Shillaker (changes requested):
Awesome, it'll be great once this is in. Just a couple of comments, the main one being about potentially removing the interpreter mode, but let's see how the offline discussion goes.

include/sgx/faasm_sgx_native_symbols_wrapper.h
csegarragonz: As discussed offline, I'll do this in a separate PR.

CMakeLists.txt
csegarragonz: Agree. Will do it in a separate PR.

tests/utils/worker_utils.cpp
csegarragonz: You are right. I will delete this.

src/wamr/CMakeLists.txt
```diff
@@ -5,40 +5,30 @@
 # We have our own custom CMake set-up which looks a bit like the wamrc build:
 # https://github.com/bytecodealliance/wasm-micro-runtime/blob/main/wamr-compiler/CMakeLists.txt
 
-# Turn on WAMR features
```
csegarragonz, line 8: @Shillaker this has changed since your first review.

In short, adding the stricter spec checks globally made the compilation here fail. I took the chance to then clean the wamr build the same way it is done for SGX.

In the process, I realised that we are relying on `LIBC_WASI` for `WAMR` (non-SGX) builds. Removing this dependency means that we need to add quite some function definitions (like `fd_close`, `fd_seek`, `fd_write`, ...). I defer this to a future PR, but soon enough as it will easily fall in oblivion.

Shillaker: 👍

Shillaker (approved):

PR #445, opened 2021-06-29
Add codegen for SGX

csegarragonz:
WAMR codegen needs [two additional parameters to generate AoT code for SGX](https://github.com/bytecodealliance/wasm-micro-runtime/blob/cba4c782735becf98f9aec9e60bd049e24620b85/wamr-compiler/main.c#L195). This PR includes the changes necessary to do this.

Now, machine code may be generated from the command line using:
```
(faasm) inv codegen <user> <function> [--wamr] [--sgx]
```

Similarly, we can invoke the function runner using:
```
(faasm) inv run <user> <function> [--wamr] [--sgx]
```

If the user provides the `--sgx` flag without the `--wamr` one, an error is prompted. It may be a bit more verbose, but I think we need to clearly state that you can't run SGX without WAMR.

Note that running SGX functions still only works in development clusters. For running in a production one, there are more changes to be made in the upload process. This is part of the breakdown of the attestation PR.

### DIFF
--- a/faasmcli/faasmcli/tasks/codegen.py
+++ b/faasmcli/faasmcli/tasks/codegen.py
@@ -28,7 +28,7 @@ def codegen(ctx, user, function, wamr=False, sgx=False):
         }
     )
 
-    if ((not wamr) and sgx):
+    if (not wamr) and sgx:
         print("Can't run SGX codegen with WAVM. Add --wamr flag.")
         exit(1)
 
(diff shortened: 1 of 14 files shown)


faasmcli/faasmcli/util/codegen.py
```diff
@@ -5,8 +5,5 @@ def find_codegen_shared_lib():
     return find_command("codegen_shared_obj")
 
 
-def find_codegen_func(wamr=False):
```
csegarragonz, line 8: This bit of code was not used anymore

src/storage/FileLoader.cpp
```diff
@@ -38,17 +38,19 @@ std::string FileLoader::getHashFilePath(const std::string& path)
...
     if (conf.wasmVm == "wamr") {
 #if (WAMR_EXECTION_MODE_INTERP)
         throw std::runtime_error(
           "WAMR codegen not supported with WAMR interp mode");
 #else
-        return wasm::wamrCodegen(bytes);
+        return wasm::wamrCodegen(bytes, isSgx);
 #endif
     } else {
+        assert(isSgx == false);
```
csegarragonz, line 53: We already check this in the same file (lines 74-76) so I thought an assertion would suffice here.

faasmcli/faasmcli/tasks/codegen.py
```diff
@@ -27,31 +28,35 @@ def codegen(ctx, user, function, wamr=False):
...
-        "{} {} {}".format(binary, user, function),
+        "{} {} {} {}".format(binary, user, function, "--sgx" if sgx else ""),
         shell=True,
         env=env,
         check=True,
     )
 
 
 @task
-def user(ctx, user, wamr=False):
```
csegarragonz, line 40: These optional parameters were really never used, and they would not work neither.

Shillaker (approved):
Nice one 👌

PR #448, opened 2021-06-30
Chained SGX function calls

csegarragonz:
This PR provides support for function chaining in SGX.

Given the way we are planning on doing attestation, the only chaining supported is "chain by name", where a user chains two registered functions (`userA/funcA -> userA/funcB`).

The alternative "chain by pointer" nesting model could be problematic with enclaves.

### DIFF
--- a/src/sgx/faasm_sgx_enclave.edl
+++ b/src/sgx/faasm_sgx_enclave.edl
@@ -21,47 +21,151 @@ enclave{
 #endif
 
     trusted{
-        public faasm_sgx_status_t faasm_sgx_enclave_unload_module(uint32_t thread_id);
-        public faasm_sgx_status_t faasm_sgx_enclave_call_function(uint32_t thread_id, uint32_t func_id);
+        public faasm_sgx_status_t faasm_sgx_enclave_unload_module(
+            uint32_t thread_id
+        );
+
+        public faasm_sgx_status_t faasm_sgx_enclave_call_function(
+            uint32_t thread_id
+        );
+
 #if(FAASM_SGX_ATTESTATION)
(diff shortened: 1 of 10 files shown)


src/sgx/faasm_sgx_enclave.edl
```diff
@@ -21,47 +21,151 @@ enclave{
 #endif
```
csegarragonz, line 21: This file was horrible to read before, and doesn't get formatted, so manually formatted it myself. I also add comments in the relevant parts that have changed in this PR.

src/sgx/faasm_sgx_enclave.edl
```diff
@@ -21,47 +21,151 @@ enclave{
...
 
     trusted{
-        public faasm_sgx_status_t faasm_sgx_enclave_unload_module(uint32_t thread_id);
-        public faasm_sgx_status_t faasm_sgx_enclave_call_function(uint32_t thread_id, uint32_t func_id);
+        public faasm_sgx_status_t faasm_sgx_enclave_unload_module(
+            uint32_t thread_id
+        );
+
+        public faasm_sgx_status_t faasm_sgx_enclave_call_function(
+            uint32_t thread_id
```
csegarragonz, line 29: No function pointer here.

src/sgx/faasm_sgx_enclave.edl
```diff
@@ -21,47 +21,151 @@ enclave{
...
+        void ocall_faasm_write_output(
+            [in, size=output_size]  uint8_t* output,
+                                    unsigned int output_size
+        );
+
+        unsigned int ocall_faasm_chain_name(
+            [in, string]            const char* name,
+            [in, size=input_size]   uint8_t* input,
+                                    long input_size
+        ) allow(faasm_sgx_enclave_load_module);
```
csegarragonz, line 153: If an `ecall` is to be called from an `ocall` (like loading a module when chaining a function), we need to add it in this `allow()` list.

Shillaker (approved):
This looks great, nice job doing the tidying of the `.edl` file too.

csegarragonz:
CC @dgoltzsche

PR #449, opened 2021-07-02
Crypto module for SGX

csegarragonz:
This PR introduces a small API to do symmetric encryption inside an enclave. Trying to run the same library outside an enclave proved tricky and will never happen at runtime, so we don't do it for the moment.

Additionally, this PR introduces a mechanism to test enclave logic from the native test suite. I add a simple test to check that an enclave can be initialised, and another one checking the correct behaviour of the symmetric encryption primitives.


include/sgx/crypto/utils.h
```diff
@@ -0,0 +1,43 @@
...
+    FaasmSgxMac_t mac;
+} FaasmSgxEncryptedMsg_t;
+
+extern "C"
+{
+    // --------------------------------------
+    // Symmetric encryption
+    // --------------------------------------
+
+    FaasmSgxEncryptedMsg_t* doSymEncrypt(FaasmSgxMsg_t* decryptedMsg,
```
csegarragonz, line 32: I add an additional level of indirection so that the actual algorithm of choice is transparent to enclave code, and so that we could potentially support different algorithms.

The mechanism to actually choose that algorithm through the config is not implemented here.

src/sgx/CMakeLists.txt
```diff
@@ -62,14 +62,77 @@ set(FAASM_SGX_WAMR_SRC_UNTRUSTED
         ${PLATFORM_SHARED_SOURCE_UNTRUSTED}
         )
 
+# --------------------------------------------------------
```
csegarragonz, line 65: Re-ordered a bit some of the targets declared here. Not a proper re-factor, but otherwise it was very difficult to track dependencies.

I'd really like to do a more in-depth re-factor of this file though.

src/sgx/CMakeLists.txt
```diff
@@ -62,14 +62,77 @@ set(FAASM_SGX_WAMR_SRC_UNTRUSTED
...
+set(FAASM_SGX_ENCLAVE_C_FLAGS ${SGX_C_GLOBAL_FLAGS}
+        -nostdinc
+        -fvisibility=hidden
+        -fpie
+        -ffunction-sections
+        -fdata-sections
+        -fstack-protector-strong
+        )
+
+# --------------------------------------------------------
```
csegarragonz, line 122: Change 1/2: added this submodule.

src/sgx/faasm_sgx_system.cpp
```diff
@@ -16,6 +16,11 @@ sgx_enclave_id_t globalEnclaveId;
 
 namespace sgx {
 
+sgx_enclave_id_t getGlobalEnclaveId()
```
csegarragonz, line 19: This was declared in the header, but never implemented.

src/sgx/faasm_sgx_checks.cpp
```diff
@@ -0,0 +1,41 @@
...
+#include <tlibc/string.h>
+
+#include <sgx/crypto/utils.h>
+#include <sgx/faasm_sgx_error.h>
+
+extern "C"
+{
+    faasm_sgx_status_t faasm_sgx_enclave_crypto_checks(void)
+    {
+        // Test string of size 13 (including trailing nullptr)
```
csegarragonz, line 12: This is the actual encryption test.

src/sgx/faasm_sgx_system.cpp
```diff
@@ -170,6 +201,10 @@ std::string faasmSgxErrorString(faasm_sgx_status_t status)
         ERROR_PRINT_CASE(FAASM_SGX_INVALID_PAYLOAD_LEN)
         ERROR_PRINT_CASE(FAASM_SGX_FUNCTION_NOT_WHITELISTED)
         ERROR_PRINT_CASE(FAASM_SGX_WAMR_FUNCTION_NOT_IMPLEMENTED)
+        // TODO - not do this manually
```
csegarragonz, line 204: For each error defined in `faasm_sgx_error.h`, we need to add an entry here so that the error text is printed. Adding a `TODO` so that we register these differently in the future.

include/sgx/crypto/types.h
```diff
@@ -0,0 +1,21 @@
...
+
+typedef uint8_t FaasmSgxIv_t[SGX_AESGCM_IV_SIZE];
+typedef uint8_t FaasmSgxMac_t[SGX_AESGCM_MAC_SIZE];
+typedef uint8_t FaasmSgxSymKey_t[SGX_AESGCM_KEY_SIZE];
+
+typedef struct FaasmSgxMsg_t
+{
+    uint8_t* buffer;
+    size_t bufferLen;
+} FaasmSgxMsg_t;
```
Shillaker, line 13: Can we avoid repeating the name and the `_t` suffix, i.e.?

```
struct FaasmSgxMsg 
{
    uint8_t* buffer;
    size_t bufferLen;
}
```

Shillaker (approved):
Nice one, LGTM other than the tiny `struct` syntax question which you can take or leave.

PR #455, opened 2021-07-23
Use only S3 storage

Shillaker:
This PR follows on from #432 and removes the optionality around the file storage backend, using only S3 (Minio in a local environment).

The file upload/ access part of the codebase was originally hacked together in a rush, and hasn't seen any updates for a long time. As a result, updating it revealed quite a bit of flaky/ unstable code as well as big gaps in testing. I've tried to add missing tests and improve error handling and logging where possible. 

Changes:

- Merge the `S3FileLoader` with the `FileLoader` interface to leave one concrete `FileLoader` class. 
- Merge everything from `conf/function_utils` into the `FileLoader` as well (as this was the only place it was used).
- Remove `LocalFileLoader` and `FileserverFileLoader` as these are no longer needed.
- Remove distinction between `shared` and `shared_store` directories which were only needed to support the fileserver (now we just have a single `shared` directory for caching shared files locally).
- Group functions related to the same file types in the `FileLoader` class (function wasm files, function object files, shared files etc.).
- Add new tests for `FileLoader` and repurpose all useful `FileserverFileLoader` and `LocalFileLoader` tests.
- Move codegen into a separate `codegen/MachineCodeGenerator` class which _uses_ the `FileLoader`, rather than having the codegen be _part of_ the `FileLoader`.
- Add Python CLI to the local dev cluster setup.
- Add tests for distributed Python file handling and upload server error handling.
- Make the distributed tests more representative of a real deployment by adding the CPP client to compile and upload functions to the upload server (rather than relying on shared directories) 
- Remove unnecessary `if (boost::filesystem::exists(path))` checks before `boost::filesystem::remove(path)`.

Note that both the `FileLoader` and `UploadServer` classes have basically been rewritten, so the diff isn't very useful.


src/wamr/codegen.cpp
```diff
@@ -75,6 +75,10 @@ std::vector<uint8_t> wamrCodegen(std::vector<uint8_t>& wasmBytes, bool isSgx)
     // Read object bytes back in
     std::vector<uint8_t> objBytes =
       faabric::util::readFileToBytes(temp.c_str());
+
+    // Delete the file
+    boost::filesystem::remove(temp.c_str());
```
Shillaker, line 80: This avoids all the WAMR codegen temporary files hanging around.

docs/api.md
```diff
@@ -3,12 +3,11 @@
 Most interactions with Faasm can be performed through the [CLI](cli.md),
```
Shillaker, line 3: Just trimming/ formatting in this file

tests/test/storage/test_shared_files.cpp
```diff
@@ -1,32 +1,47 @@
...
 #include <faabric/util/files.h>
+#include <storage/FileLoader.h>
 #include <storage/SharedFiles.h>
 
 using namespace storage;
 
 namespace tests {
-TEST_CASE("Check sync shared file", "[storage]")
+
+class SharedFilesTestFixture : public S3TestFixture
```
csegarragonz, line 18: Why isn't this under `utils/faasm_fixtures.h`? What is the criterion to include fixtures or not?

I assume we include fixtures there if they are used in more than one file, that works for me, but just double-checking.

csegarragonz (changes requested):
LGTM. Couple questions and minor-comments.

include/upload/UploadServer.h
```diff
@@ -25,23 +28,29 @@ class UploadServer
...
-
-    static std::vector<std::string> getPathParts(const http_request& request);
-
-    static std::vector<uint8_t> getState(const http_request& request);
-
   private:
     bool stopped = false;
 
-    static void handlePythonFunctionUpload(const http_request& request);
+    static std::vector<uint8_t> getState(const std::string& user,
```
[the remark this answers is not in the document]
Shillaker: Remember that `static` here means a `static` class variable (not `static` in the sense of global variables/ functions), so I'm not sure what relevance its `private` status has (i.e. you can have `public` static class methods too). The reason these methods are `static` is because the `cpprestsdk` requires that requests are handled by using static methods (e.g. [`handleGet`](https://github.com/faasm/faasm/blob/master/include/upload/UploadServer.h#L22)), and these methods are called from within those.

src/codegen/MachineCodeGenerator.cpp
```diff
@@ -0,0 +1,144 @@
+#if (WAMR_EXECUTION_MODE_INTERP)
```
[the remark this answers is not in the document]
Shillaker: No, this is just ported over from where it lived before. If we no longer support interpreter mode we should remove `WAMR_EXECUTION_MODE_INTERP` completely (it still exists in `CMakeLists.txt`, `WAMRWasmModule.cpp` etc.). Removing it from here would be part of that work.

tests/test/codegen/test_machine_code_generator.cpp
```diff
@@ -0,0 +1,232 @@
...
+
+    bool isCodegenRepeatable = true;
+    SECTION("WAVM codegen")
+    {
+        conf.wasmVm = "wavm";
+        objectFileA = "/tmp/obj/demo/hello/function.wasm.o";
+        objectFileB = "/tmp/obj/demo/echo/function.wasm.o";
+    }
+
+    SECTION("WAMR codegen")
```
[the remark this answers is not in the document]
Shillaker: I've added this now but it seems the SGX codegen is even less repeatable than the normal WAMR one. Normal WAMR produces the same number of bytes but with different contents, whereas the SGX codegen produces slightly different lengths of bytes each time 🤷 . I guess we'll learn more as we get into the SGX experiments in more detail.

tests/test/storage/test_shared_files.cpp
Shillaker: Yes it's not a hard and fast rule but generally fixtures should contain things that are useful to testing more than one _component_ of the system (this is _mostly_ indicated by being needed in more than one test file). In this case, this logic is only needed for testing the shared files, so doesn't merit a fixture. However, the other fixture in this PR where it reads in wasm and object files up front, is useful for testing machine code generation and shared file handling, so is in a fixture. 

All that said, we do have to be careful to keep the fixtures simple, and make sure we're not testing the same system component in multiple files (as this suggests wasted effort or a poor class/ module structure). Again though, no hard and fast rules, just make a call case-by-case.

csegarragonz (approved):

PR #458, opened 2021-08-03
Improve shared files test

Shillaker:
Noticed that this test wasn't checking the loading of files from S3, so I updated it.

### DIFF
--- a/tests/test/storage/test_shared_files.cpp
+++ b/tests/test/storage/test_shared_files.cpp
@@ -1,6 +1,5 @@
 #include <catch2/catch.hpp>
 
-#include "faabric/util/func.h"
 #include "faasm_fixtures.h"
 #include "utils.h"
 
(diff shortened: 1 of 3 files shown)


tests/test/storage/test_shared_files.cpp
```diff
@@ -1,35 +1,19 @@
...
-      : loader(storage::getFileLoader())
-    {
-        SharedFiles::clear();
-    }
-
-    ~SharedFilesTestFixture() { SharedFiles::clear(); }
-
-  protected:
-    storage::FileLoader& loader;
-};
```
Shillaker, line 31: As discussed in previous PR, this part of the system is now involved in tests in more than one file, so can be promoted into `faasm_fixtures.h`.

csegarragonz (approved):

PR #460, opened 2021-08-10
Add test for FFmpeg

csegarragonz:
Add support for running FFmpeg tests in Faasm, and bump the code version.


tests/test/ported_libs/test_ffmpeg.cpp
```diff
@@ -0,0 +1,86 @@
...
+#include <iterator>
+#include <vector>
+
+#include <storage/FileLoader.h>
+#include <storage/SharedFiles.h>
+#include <upload/UploadServer.h>
+
+namespace tests {
+
+class FileUploadTestFixture
```
csegarragonz, line 16: Other fixtures did more things than I needed. Namely, all `S3TestFixture` descendants use a separate bucket `faasm-test`.

If we use this here, it means we need to re-run the code generation at test runtime, which adds minutes of computations.

In this fixture I only care about uploading (and clearing) shared files. I think this may be useful for other library-related tests.

Shillaker: I think this fixture is doing too much, why does it need to go through the full upload process with the shared file? It's pulling in other unnecessary components that are tested elsewhere, and makes this quite a big fixture. Have a look at [`test_shared_files.cpp`](https://github.com/faasm/faasm/blob/master/tests/test/faaslet/test_shared_files.cpp), this keeps it more simple.

Using a separate bucket is good because it isolates tests from messing up someone's local dev environment. This is also very important wrt shared files, which hold all the python functions used by other tests (so we don't want to delete that folder in other tests). It's not immediately clear why we can't do this here, I would have thought that even if the bucket was changed, if the machine code and shared file were present on the local filesystem it would just load them from there. However, if it is necessary, I would put a note about the codegen/ file size as a comment too.

I would rename this to `FFmpeg` test fixture, that would look something like this (and combined with comments below):

```cpp
// NOTE: we must be careful with this fixture that we don't have to rerun the codegen for 
// the ffmpeg function as it's over 40MB of wasm and takes several minutes.
class FFmpegTestFixture {
public:
    FFmpegTestFixture() 
        : loader(storage::getFileLoader()) {
        storage::SharedFiles::clear();
        
        // Remove the file
        filePath = loader.getSharedFileFile(relativeSharedFilePath);
        boost::filesystem::remove(filePath);

        // Upload the file
        fileBytes = faabric::util::readFileToBytes(testFilePath);
        loader.uploadSharedFile(relativeSharedFilePath, fileBytes);
    }
    
    ~FFMpegTestFixture() {
          storage::SharedFiles::clear();
          
          boost::filesystem::remove(filePath);
    }
protected:
    storage::FileLoader& loader;
    
    std::string testFilePath = "files/ffmpeg_video.mp4";
    std::string filePath;
    std::string relativeSharedFilePath = "ffmpeg/sample_video.mp4";
    std::vector<uint8_t> fileBytes;
};

TEST_CASE_METHOD(FFmpegTestFixture,
                 "Test executing FFmpeg checks",
                 "[libs]")
{
    faabric::Message msg = faabric::util::messageFactory("ffmpeg", "check");
    executeFunction(msg);
}
```

Shillaker (changes requested):

Shillaker (approved):

PR #472, opened 2021-09-02
WAMR WASI Argc/argv and Filesystem Implementation

csegarragonz:
This PR was originally meant just to implement WASI support for FFmpeg to run on WAMR, but has ended up including many different changes in the WAMR-related source code.
Change log:
* Bump our WAMR fork to a recent upstream commit.
* Re-write our WAMR CMake file to use the CMake config flags WAMR defines for projects linking against it (I did not manage to re-write the CMake file without bumping the WAMR fork to an up-to-date commit).
* Fix WAMR codegen bug in which SGX codegen was stored under non-SGX WAMR machine code. This fix, together with the CMake + commit bump mean that now WAMR and WAMR+SGX generate consistent object code (just the bug fix with the current master would not work).
* Implement the relevant WASI system calls needed for FFmpeg and the corresponding tests. This means that almost all tests in `faaslet/test_filesystem.cpp`, `faaslet/test_env.cpp` and `faaslet/test_io.cpp` are covered. Providing full-WASI support for WAMR after this PR should not be much more work.
* In a "i don't understand why this is not working" moment I bumped the code version to see if that would fix the tests. In fact this was not really needed, the problem stemming from the WAMR codegen bug, in which code that I thought was being generated in GHA in reality it was not, as the system thought it had it cached.

There are some features that I'd like to address in a separate PR, as this is already quite big.
Namely:
* Full-support for the remaining filesystem, env, and io WASI calls.
* The last point includes fixing the `fdread` system call that exposed a bug in WAMR's WASI build that I am [currently upstreaming](https://github.com/bytecodealliance/wasm-micro-runtime/pull/766).


src/wamr/pthread.cpp
```diff
@@ -23,20 +49,55 @@ static int32_t pthread_mutex_init_wrapper(wasm_exec_env_t exec_env,
 
 static int32_t pthread_mutex_lock_wrapper(wasm_exec_env_t exec_env, int32_t a)
 {
-    SPDLOG_DEBUG("S - pthread_mutex_lock {}", a);
+    SPDLOG_TRACE("S - pthread_mutex_lock {}", a);
```
csegarragonz, line 52: Switched `lock` and `unlock` to `TRACE` as they cluttered outputs a lot and, given that they are no-opped, I figured they did not matter that much (for the moment).

src/wasm/WasmModule.cpp
```diff
@@ -227,6 +227,11 @@ uint32_t WasmModule::getArgc()
     return argc;
 }
 
+std::vector<std::string> WasmModule::getArgv()
+{
+    return argv;
```
csegarragonz, line 232: This is the only change I need in the `WasmModule` base class. It is used in the `args_get` system call.

The reason this is not needed in the WAVM implementation is because we use `WasmModule::argv` which we call from [`WAVMWasmModule::writeArgvToMemory`](https://github.com/faasm/faasm/blob/d60d2a8d42842085bd6b5d8961a54fdcf6ee73ba/src/wavm/WAVMWasmModule.cpp#L499). IMO the approach in WAVM adds unnecessary complexity. This means WAVM and WAMR's implementation slightly differ. I am happy to change either if you think it preferrable.

csegarragonz:
Re-based after merging #471 (so tests are running again)

src/wasm/WasmModule.cpp
Shillaker: `argv` is not private, it's `protected` (so that it can be accessed directly in a subclass). In general I actually think it's better to push as much logic as we can into the `XYZWasmModule` classes, hence it's all contained in one place, not scattered over the intrinsic/ native call implementations, and is in theory more reusable.

src/wamr/filesystem.cpp
```diff
@@ -1,20 +1,432 @@
...
+    std::copy(pathStr.begin(), pathStr.end(), path);
+
+    return __WASI_ESUCCESS;
+}
+
+typedef struct wasi_prestat_app
+{
+    __wasi_preopentype_t pr_type;
+    uint32_t pr_name_len;
+} wasi_prestat_app_t;
```
Shillaker, line 152: I think we normally declare structs like this:

```
struct wasi_prestat_app_t
{
    __wasi_preopentype_t pr_type;
    uint32_t pr_name_len;
};
```

src/wamr/filesystem.cpp
```diff
@@ -1,20 +1,432 @@
...
+
+static int32_t wasi_fd_prestat_get(wasm_exec_env_t exec_env,
+                                   int32_t fd,
+                                   wasi_prestat_app_t* prestatApp)
+{
+    SPDLOG_DEBUG("S - fd_prestat_get {}", fd);
+
+    WAMRWasmModule* module = getExecutingWAMRModule();
+    if (!module->getFileSystem().fileDescriptorExists(fd)) {
```
Shillaker, line 161: Same point re. calling `getFileSystem()` twice and not using `module` variable.

src/wamr/filesystem.cpp
```diff
@@ -1,20 +1,432 @@
...
+                              int32_t openFlags,
+                              int64_t rightsBase,
+                              int64_t rightsInheriting,
+                              int32_t fdFlags,
+                              int32_t* fdApp)
+{
+    WAMRWasmModule* module = getExecutingWAMRModule();
+
+    VALIDATE_NATIVE_ADDR(path, pathLen);
+    const std::string pathStr(path, pathLen);
```
Shillaker, line 354: This is an example of where I'm not sure about the use of `VALIDATE_NATIVE_ADDR`, first of all I don't think we do these sort of checks on strings passed to WAMR natives elsewhere (as I'm not sure they're necessary), but more importantly returning an `EBADF` code doesn't seem right when the path passed to `path_open` is invalid (as it's not related to a file descriptor), so could cause quite confusing issues.

src/wamr/filesystem.cpp
```diff
@@ -1,20 +1,432 @@
...
+
+    return wasiErrno;
+}
+
+static int32_t wasi_fd_write(wasm_exec_env_t exec_env,
+                             int32_t fd,
+                             iovec_app_t* ioVecBuffApp,
+                             int32_t ioVecCountApp,
+                             int32_t* bytesWritten)
+{
```
Shillaker, line 254: This method looks incredibly similar to the WAVM version, and it seems like there's not much WAMR/ WAVM specific stuff in it, could we capture some of the logic in the `FileSystem` class instead to avoid duplication?

src/wamr/filesystem.cpp
```diff
@@ -1,20 +1,432 @@
...
+    prestatApp->pr_name_len = fileDesc.getPath().size();
+
+    return __WASI_ESUCCESS;
+}
+
+typedef struct iovec_app
+{
+    uint32_t buffOffset;
+    uint32_t buffLen;
+} iovec_app_t;
```
Shillaker, line 180: There are a few inlined struct definitions in this file that are easy to miss, and some are already defined in the WAVM code so it seems a shame to duplicate them. However, the WAVM-based versions rely on other definitions in the WAVM headers so we can't really mix the two. Either way, it would be good to move all the struct definitions from in here to a separate header file as we have in the WAVM-based version.

src/wamr/filesystem.cpp
```diff
@@ -1,20 +1,432 @@
...
+        ioVecBuffNative[i] = ioVecNative;
+    }
+
+    // Read from fd
+    VALIDATE_NATIVE_ADDR(bytesRead, sizeof(int32_t))
+    *bytesRead = ::readv(fileDesc.getLinuxFd(), ioVecBuffNative, ioVecCountApp);
+
+    delete[] ioVecBuffNative;
+
+    return __WASI_ESUCCESS;
```
Shillaker, line 216: Same comment about duplicating the logic for this function from the existing WAVM implementation, could we refactor it into the `FileSystem` class?

Shillaker (changes requested):
A few changes listed in the code, the main issue being the duplication from the WAVM version and lack of more granular tests.

tests/test/ported_libs/test_ffmpeg.cpp
```diff
@@ -58,4 +58,11 @@ TEST_CASE_METHOD(FFmpegTestFixture,
     faabric::Message msg = faabric::util::messageFactory("ffmpeg", "check");
     execFunction(msg);
 }
+
+TEST_CASE_METHOD(FFmpegTestFixture,
+                 "Test executing FFmpeg checks in WAMR",
+                 "[libs]")
+{
+    executeWithWamrPool("ffmpeg", "check");
+}
```
Shillaker, line 67: This shouldn't be the only test that exercises the WAMR filesystem syscalls, instead we can think about reusing the wasm functions already executed using the WAVM-based version in `tests/test/faaslet/test_filesystem.cpp`.

include/wamr/WAMRWasmModule.h
```diff
@@ -45,6 +39,14 @@ class WAMRWasmModule final : public WasmModule
     // ----- Module lifecycle -----
     void doBindToFunction(faabric::Message& msg, bool cache) override;
 
+    void writeStringArrayToMemory(const std::vector<std::string>& strings,
```
csegarragonz, line 42: Unfortunately, I could not re-use the prototypes in `WasmModule.h` as the parameter types did not match. IMO I'd delete those from there, and leave two different implementations in each `XYZWasmModule` but up for discussion.

src/wamr/filesystem.cpp
```diff
@@ -1,20 +1,432 @@
...
+{
+    SPDLOG_DEBUG("S - fd_prestat_dir_name {}", fd);
+
+    WAMRWasmModule* module = getExecutingWAMRModule();
+    if (!module->getFileSystem().fileDescriptorExists(fd)) {
+        return __WASI_EBADF;
+    }
+
+    storage::FileDescriptor& fileDesc =
+      module->getFileSystem().getFileDescriptor(fd);
```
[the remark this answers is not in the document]
csegarragonz: In this case I will keep the `WAMRWasmModule* module` variable, as I then need to call `module->writeStringToWasmMemory`.

include/wamr/WAMRWasmModule.h
```diff
@@ -7,6 +7,26 @@
 #define STACK_SIZE_KB 8192
 #define HEAP_SIZE_KB 8192
 
+// Address translation macros
+
+#define VALIDATE_NATIVE_ADDR(ptr, size)                                        \
```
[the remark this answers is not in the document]
csegarragonz: `VALIDATE_NATIVE_ADDR` was a macro wrapping the usage of `wasm_runtime_validate_native_address(wasm_module_t, native_ptr_t, size_t)` that validates that a given address belongs to the given module.

I agree that the implemented macro:
1. Shadowed the depdendency to a particular `WAMRWasmModule`
2. Returned an incorrect error code

Alternatively, I have now:
1. Implemented a class method on `WAMRWasmModule` to validate a native address
2. Throw a runtime exception

Even further we could:
1. No-op the check for `Release` builds

src/wamr/filesystem.cpp
csegarragonz: I need the `module` variable here to call `module->validateNativeAddress`.

src/wamr/filesystem.cpp
csegarragonz: See the reply on the main `VALIDATE_NATIVE_ADDR` thread to see the points I make there.

faasmcli/faasmcli/tasks/codegen.py
```diff
@@ -93,8 +93,27 @@ def local(ctx):
...
+    codegen(ctx, "demo", "argc_argv_test", wamr=True)
+    codegen(ctx, "demo", "exit", wamr=True)
+    codegen(ctx, "demo", "getenv", wamr=True)
+
+    # Filesystem
+    codegen(ctx, "demo", "fcntl", wamr=True)
+    codegen(ctx, "demo", "file", wamr=True)
+    codegen(ctx, "demo", "filedescriptor", wamr=True)
+    codegen(ctx, "demo", "fstat", wamr=True)
+    codegen(ctx, "demo", "fread", wamr=True)  # not working
```
csegarragonz, line 110: Hopefully I can upstream the changes to get this to work: bytecodealliance/wasm-micro-runtime#766

src/wamr/CMakeLists.txt
```diff
@@ -1,90 +1,86 @@
...
-set(ALL_WAMR_SRC ${COMMON_WAMR_SRC} ${EXTRA_WAMR_SRC})
-faasm_private_lib(wamrlib "${ALL_WAMR_SRC}")
+# Explicitely set to 0. If set to 1, WAMR will allocate the module's instance
+# memory using `malloc` instead of `mmap` and calls to `mprotect` fail.
+set (WAMR_DISABLE_HW_BOUND_CHECK 0)
+# This definition prevents a buffer underflow during code generation
+add_definitions(-DWASM_ENABLE_WAMR_COMPILER=1)
+
+include (${WAMR_ROOT_DIR}/build-scripts/runtime_lib.cmake)
+# 13/09/2021 - Setting the CMake flag includes the code required to run
```
csegarragonz, line 35: The criteria for having a date here and nowhere else, is that the `remove_definitions` bit is a patch, and the other are configuration flags.

Nonetheless, I also add for each configuration flag what happens if set otherwise. Strikingly enough, most flags will break the build if set to the opposite value.

tests/test/codegen/test_machine_code_generator.cpp
```diff
@@ -79,8 +79,6 @@ TEST_CASE_METHOD(CodegenTestFixture,
     std::string objectFileA;
     std::string objectFileB;
 
-    bool isCodegenExactlySame = true;
```
csegarragonz, line 82: Thanks to the WAMR upgrade we can now remove these checks :partying_face:

tests/test/codegen/test_machine_code_generator.cpp
```diff
@@ -248,4 +231,76 @@ TEST_CASE_METHOD(CodegenTestFixture,
       loader.loadSharedObjectObjectHash(localSharedObjFile);
     REQUIRE(hashAfter == hashBefore);
 }
+
+TEST_CASE_METHOD(CodegenTestFixture,
```
csegarragonz, line 235: Before, running this combination would only generate code once, and confuse the cached version of one for the other one.

tests/test/wamr/test_wamr.cpp
```diff
@@ -55,6 +55,6 @@ TEST_CASE("Test WAMR sbrk", "[wamr]")
 
 TEST_CASE("Test executing chain function with WAMR", "[wamr]")
 {
-    executeWithWamrPool("demo", "chain");
+    executeWithWamrPool("demo", "chain", 10000);
```
csegarragonz, line 58: This test would sometimes time out, so adding a longer timeout

.github/workflows/tests.yml
```diff
@@ -160,12 +160,6 @@ jobs:
         run: echo "CPU_MODEL=$(./bin/print_cpu.sh)" >> $GITHUB_ENV
       - name: "Print CPU model"
         run: echo "${{ env.CPU_MODEL}}"
-      # --- Cache ---
```
csegarragonz, line 163: Unfortunately, WAMR machine code generated with one version of the system might not be compatible with other versions. Consequently, tests for which machine code was cached would fail for no aparent reason.

I know this makes test runs even slower that they currently are. I suggest we reintroduce the cache when the WAMR/SGX build and code is consolidated.

src/wamr/filesystem.cpp
csegarragonz: This PR is already very cluttered so I suggest I do this in a separate PR.
Of course, the argument goes both ways, and given how cluttered this PR is I could also do it here.

In favour of the former, there are a couple more `TODO`s in the main PR description that I must do regardless, so it's not likely that I will forget about this at all.

src/wamr/WAMRWasmModule.cpp
```diff
@@ -187,29 +199,139 @@ int WAMRWasmModule::executeWasmFunction(const std::string& funcName)
     } else {
         std::string errorMessage(
           ((AOTModuleInstance*)moduleInstance)->cur_exception);
-        SPDLOG_ERROR("Function failed: {}", errorMessage);
+        SPDLOG_ERROR("Caught wasm runtime exception: {}", errorMessage);
+
+        // TODO - agree on a protocol to send the return code through the error
+        // message (throwing exceptions not supported). For the moment, if the
+        // error message is a "Exception: 0" we assume no errors (i.e. exit(0))
```
Shillaker, line 206: Is the `Exception: 0` string defined somewhere in WAMR or is that something we've created? Can we link to the WAMR constant if so? If not, we should create our own constant which should live in `#define WAMR_AOT_FAIL_MESSAGE "Exception: 0"` or something like that.

What is the scenario where the result of `aot_create_exec_env_and_call_function` is `false`, yet we're returning `0`, implying that the function was a success? Surely whenever `aot_create_exec_env_and_call_function` returns `false` we should at least return `1` (if not the actual error code).

If we control the format of this error message, it makes most sense to me just to have a string of the return code, then call `stoi` on it.

src/wamr/WAMRWasmModule.cpp
```diff
@@ -187,29 +199,139 @@ int WAMRWasmModule::executeWasmFunction(const std::string& funcName)
...
         return 1;
     }
 
     return 0;
 }
 
+void WAMRWasmModule::writeStringToWasmMemory(const std::string& strHost,
+                                             char* strWasm)
+{
+    this->validateNativeAddress(strWasm, strHost.size());
```
Shillaker, line 220: We don't use `this->` unless it's absolutely necessary to avoid collisions between function names (and even then we should reconsider renaming the function).

src/wamr/WAMRWasmModule.cpp
```diff
@@ -187,29 +199,139 @@ int WAMRWasmModule::executeWasmFunction(const std::string& funcName)
...
+void WAMRWasmModule::writeStringArrayToMemory(
+  const std::vector<std::string>& strings,
+  uint32_t* strOffsets,
+  char* strBuffer)
+{
+    this->validateNativeAddress(strOffsets, strings.size() * sizeof(uint32_t));
+
+    char* nextBuffer = strBuffer;
+    size_t i;
+    for (i = 0; i < strings.size(); ++i) {
```
Shillaker, line 233: We define loop variables inline, i.e. 

```
for(size_t i = 0; i < strings.size(); ++i) {
```

src/wamr/WAMRWasmModule.cpp
```diff
@@ -187,29 +199,139 @@ int WAMRWasmModule::executeWasmFunction(const std::string& funcName)
...
+{
+    writeStringArrayToMemory(
+      wasmEnvironment.getVars(), envOffsetsWasm, envBuffWasm);
+}
+
+void WAMRWasmModule::validateNativeAddress(void* nativePtr, size_t size)
+{
+    if (!wasm_runtime_validate_native_addr(moduleInstance, nativePtr, size)) {
+        SPDLOG_ERROR("Native pointer does not point to a valid address in the "
+                     "WAMR's module address space");
```
Shillaker, line 262: The way we normally approach error is to log as much detail as possible (in this case print the pointer and its size), then throw an error with a simpler generic message. If we're just doing a simple string in both cases, it's easier to define it as a separate `errorMsg` variable and have both passed to the `SPDLOG` macro and exception.

src/wamr/WAMRWasmModule.cpp
```diff
@@ -187,29 +199,139 @@ int WAMRWasmModule::executeWasmFunction(const std::string& funcName)
...
+        throw std::runtime_error("Pointer out of WAMR's module address space");
+    }
+}
+
+void WAMRWasmModule::validateWasmOffset(uint32_t wasmOffset, size_t size)
+{
+    if (!wasm_runtime_validate_app_addr(moduleInstance, wasmOffset, size)) {
+        SPDLOG_ERROR("WASM offset outside WAMR module instance memory");
```
Shillaker, line 270: Again, here we can log the `wasmOffset` and `size` as part of the error message.

src/wamr/WAMRWasmModule.cpp
```diff
@@ -187,29 +199,139 @@ int WAMRWasmModule::executeWasmFunction(const std::string& funcName)
...
+        throw std::runtime_error("Offset outside WAMR's memory");
+    }
+}
+
+uint8_t* WAMRWasmModule::wasmPointerToNative(uint32_t wasmPtr)
+{
+    void* nativePtr = wasm_runtime_addr_app_to_native(moduleInstance, wasmPtr);
+    if (nativePtr == nullptr) {
+        SPDLOG_ERROR(
+          "WASM offset {} is out of the WAMR module's address space");
```
Shillaker, line 280: This has a space for a variable in the message but is missing that variable so won't print anything at runtime.

src/wamr/WAMRWasmModule.cpp
```diff
@@ -187,29 +199,139 @@ int WAMRWasmModule::executeWasmFunction(const std::string& funcName)
...
 
-    SPDLOG_DEBUG("WAMR growing memory by {}", nBytes);
+    uint32_t oldBytes = getMemorySizeBytes();
+    uint32_t oldBrk = currentBrk;
+    uint32_t newBrk = currentBrk + nBytes;
 
-    uint32_t memBase = currentBrk;
+    if (!isWasmPageAligned(newBrk)) {
+        SPDLOG_ERROR("Growing WAMR memory by {} is not wasm page aligned",
+                     nBytes);
```
Shillaker, line 300: Can put more detail in this message, e.g. include `currentBrk` and `newBrk`.

src/wamr/env.cpp
```diff
@@ -0,0 +1,97 @@
+#include <faabric/util/logging.h>
+#include <storage/FileDescriptor.h>
+#include <wamr/WAMRWasmModule.h>
+#include <wamr/native.h>
+
+#include <stdexcept>
+
+// WAMR includes
```
Shillaker, line 8: I don't think we need these `WAMR includes` comments on all the includes. We don't do it for any other libraries. It's sufficient to just put them after a newline.

src/wamr/env.cpp
```diff
@@ -0,0 +1,97 @@
...
+}
+
+void wasi_proc_exit(wasm_exec_env_t exec_env, int32_t retCode)
+{
+    SPDLOG_DEBUG("S - proc_exit {}", retCode);
+
+    // 08/09/2021 - WAMR doesn't support throwing and catching exceptions.
+    // Alternatively, set the error message.
+    WAMRWasmModule* module = getExecutingWAMRModule();
+    char buf[32];
```
Shillaker, line 79: It's not clear where this `32` comes from, is it `sizeof(int)`? If so, can use that instead.

src/wamr/filesystem.cpp
```diff
@@ -1,20 +1,466 @@
...
+static int32_t wasi_fd_readdir(wasm_exec_env_t exec_env,
+                               int32_t a,
+                               int32_t* b,
+                               char* c,
+                               int64_t d,
+                               int32_t e)
+{
+    SPDLOG_DEBUG("S - fd_readdir");
+    throw std::runtime_error("fd_readdir not implemented");
+    return 0;
```
Shillaker, line 235: I don't think we need a `return` statement after throwing an exception.

src/wamr/filesystem.cpp
Shillaker: Ok cool, separate PR makes sense.

src/wamr/timing.cpp
```diff
@@ -0,0 +1,39 @@
...
+
+namespace wasm {
+uint32_t wasi_clock_time_get(wasm_exec_env_t exec_env,
+                             int32_t a,
+                             int64_t b,
+                             int32_t* c)
+{
+    SPDLOG_DEBUG("S - clock_time_get");
+    throw std::runtime_error("clock_time_get not implemented");
+    return 0;
```
Shillaker, line 15: Same comment re. `return` after `throw` as above, if not necessary there are several places in the PR we can remove.

tests/test/faaslet/test_filesystem.cpp
```diff
@@ -65,34 +65,55 @@ TEST_CASE_METHOD(FilesystemTestFixture, "Test listdir", "[faaslet]")
...
+// TODO - bug in WAMR's WASI implementation makes this test fail. Will uncomment
+// when fixed.
 TEST_CASE_METHOD(FilesystemTestFixture, "Test fread", "[faaslet]")
 {
     faabric::Message msg = faabric::util::messageFactory("demo", "fread");
-    execFunction(msg);
+
+    SECTION("WAVM") { execFunction(msg); }
+
+    // SECTION("WAMR") { execWamrFunction(msg); }
```
Shillaker, line 85: Ah, this relates to the `not working` comment earlier. This is the right place to have the comment and `TODO`, we can remove the `not working` comment from the Python module as that _is_ working.

tests/test/faaslet/test_io.cpp
```diff
@@ -8,28 +8,35 @@
...
 {
     cleanSystem();
     faabric::Message call = faabric::util::messageFactory("demo", "echo");
     std::string inputData = "http://www.testinput/foo.com";
     call.set_inputdata(inputData.c_str());
 
+    conf::FaasmConfig& conf = conf::getFaasmConfig();
+    SECTION("WAVM") { conf.wasmVm = "wavm"; }
+
+    SECTION("WAMR") { conf.wasmVm = "wamr"; }
```
Shillaker, line 33: These tests are now changing the config and not resetting it at the end (leaving the `wasmVm` set to `wamr` which may break other tests). 

They should be `TEST_CASE_METHOD`s using a custom fixture inheriting from `FaasmConfTestFixture`  and `FunctionExecTestFixture` (so we can potentially remove the `cleanSystem` too).

Shillaker (changes requested):

src/wamr/WAMRWasmModule.cpp
csegarragonz: In this particular case, `nativePtr` is a `void*` that can be, potentially, corrupted. I don't think printing it out is a good idea.

I do have enriched the other error messages as suggested, and maybe use an `errorMsg` for this one.

src/wamr/WAMRWasmModule.cpp
csegarragonz: There was a missunderstanding on my side, in which I mimicked the behaviour of `libc`'s `exit` rather than `wasi_proc_exit`.

Turns out that WAMR does the same I implemented but at the runtime level. I.e. WASI sets an exception with string "wasi proc exit", and, upon exit with exception, if the string is "wasi proc exit" WAMR removes said exception and returns 0.

See source here: https://github.com/bytecodealliance/wasm-micro-runtime/blob/main/core/iwasm/aot/aot_runtime.c#L874

Shillaker (approved):

PR #474, opened 2021-09-06
Add pod anti-affinity on Knative Faasm service

Shillaker:
Adds pod anti-affinity between Faasm Knative workers. This required using a YAML config file to define the Knative service, as I couldn't work out a way to specify the affinities using the `kn` client. 

Using a YAML file means that we can't dynamically generate a service name to pass to the `kn` client, so we need to revert to having a fixed `faasm-worker` service. This isn't the end of the world, but does make it a little bit harder to debug repeated redeployments of the same service. Unfortunately I can't see a way around this.

I've also added a couple of fixes for the local microk8s set-up.


deploy/k8s/minio.yml
```diff
@@ -35,7 +35,6 @@ metadata:
   name: minio
   namespace: faasm
 spec:
-  type: ClusterIP
```
Shillaker, line 38: Removed these to let the cluster decide on its appropriate default service type.

csegarragonz (approved):

PR #475, opened 2021-09-08
Single snapshot per function chain

Shillaker:
Currently OpenMP code takes a snapshot on every batch of threaded operations, however, it's more efficient to take a single snapshot on entering the main function, then delete it on the way out. This snapshot then gets periodically updated via Faabric's snapshot dirty page checks. 

This is achieved by taking a per-app snapshot the first time we fork an OpenMP thread (named to match the `appid` field in the Faabric messages. Not sure if "app" is the right term, it's really per-function-chain but "app" is shorter and what's used in Faabric). This single snapshot is used by all subsequent OpenMP threads belonging to the "app", and deleted only when the original function has finished. This is encapsulated in the `WasmModule` where all the other module snapshot logic lives.

To test this I added a stress test for the OpenMP reduce operation, which showed up a long-standing race condition/ bug in our OpenMP implementation. This was around the `__kmpc_reduce[_nowait]` and `__kmpc_end_reduce[_nowait]`. These functions are called before and after each thread performs the reduction operation on its own copy of the data, hence need to acquire and release a unique lock to ensure that no other threads are performing the reduce at the same time. This didn't actually happen before (although there is some other locking logic around the implicit barrier on the `nowait` version, and making sure the master thread only exists when the others have finished). This PR also includes a fix for that.

### DIFF
--- a/src/threads/ThreadState.cpp
+++ b/src/threads/ThreadState.cpp
@@ -160,12 +160,12 @@ int Level::getMaxThreadsAtNextLevel() const
 
 void Level::masterWait(int threadNum)
 {
+    // All threads must lock when entering this function
     FROM_MAP(nowaitMutex, std::mutex, nowaitMutexes)
+    std::unique_lock<std::mutex> lock(*nowaitMutex);
+
     FROM_MAP(nowaitCount, std::atomic<int>, nowaitCounts)
     FROM_MAP(nowaitCv, std::condition_variable, nowaitCvs)
-
-    // All threads must lock when entering this function
-    std::unique_lock<std::mutex> lock(*nowaitMutex);
 
     if (threadNum == 0) {
         // Wait until all non-master threads have finished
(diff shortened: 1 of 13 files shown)


src/wasm/WasmModule.cpp
```diff
@@ -129,11 +177,13 @@ void WasmModule::restore(const std::string& snapshotKey)
...
                      bytesRequired);
         this->growMemory(bytesRequired);
-    } else {
+    } else if (data.size < memSize) {
         size_t shrinkBy = memSize - data.size;
         SPDLOG_DEBUG("Shrinking memory by {} bytes to restore snapshot",
                      shrinkBy);
         this->shrinkMemory(shrinkBy);
+    } else {
+        SPDLOG_DEBUG("Memory already correct size for snapshot ({})", memSize);
```
Shillaker, line 186: This logging was unclear when the memory was the right size already; it would print `Shrinking memory by 0 bytes to restore snapshot`

src/threads/ThreadState.cpp
```diff
@@ -74,10 +74,11 @@ void setCurrentOpenMPLevel(
 
     std::string funcStr = faabric::util::funcToString(req);
     currentLevel = levelFromBatchRequest(req);
-    SPDLOG_TRACE("Deserialised from {} bytes for {}, {}",
-                 req->contextdata().size(),
-                 funcStr,
-                 currentLevel->toString());
+    SPDLOG_TRACE(
+      "Deserialised thread-local OpenMP level from {} bytes for {}, {}",
```
Shillaker, line 78: This is repeated in the logs for each thread, and confusing when you don't realise it's thread-local, so added in the log.

csegarragonz (approved):
LGTM :+1:

PR #476, opened 2021-09-08
Set up config file for local cluster development

Shillaker:
Adds `--local` option to `knative.ini-file` task for running against a local cluster.


bin/cluster_dev.sh
```diff
@@ -32,6 +32,7 @@ docker-compose -f docker-compose.yml \
     up \
     --no-recreate \
     -d \
+    --scale worker=2
```
Shillaker, line 35: Wasn't actually a cluster before as it was only running one worker.

docker-compose.yml
```diff
@@ -64,6 +64,7 @@ services:
       - NETNS_MODE=off
       - REDIS_QUEUE_HOST=redis-queue
       - REDIS_STATE_HOST=redis-state
+      - OVERRIDE_CPU_COUNT=4
```
Shillaker, line 67: Setting a lower CPU count on each worker forces function executions to be distributed. Can be adjusted for smaller hosts too.

logs/worker-1-log
```diff
@@ -0,0 +1,431 @@
+[33mworker_1       |[0m Setting up cgroup
```
csegarragonz, line 1: I think these two log files accidentaly made it to the PR.

csegarragonz (changes requested):
LGTM, just make sure to delete the `logs/*` files

logs/worker-1-log
Shillaker: Ah yes, whoops!

csegarragonz (approved):

PR #482, opened 2021-09-17
Update to Network Namespaces

csegarragonz:
After different offline discussions and comments on this PR, I have ended up updating part of the network namespaces code, scripts, docs, and bindings to faasm.
In particular this PR:
* Removes the `mode` attribute of the `NetworkNamespace` class. Having both an environment variable, and a class attribute that do the same seemed unnecessary.
* Fixes race condition during network namespace initialisation.
* Fully no-ops namespace operations if the config is off.
* Introduces a new env variable `MAX_NET_NAMESPACES` to link the created namespaces, with the limits set in faasm.
* Transcribes scripts to create namespaces from bash scripts to invoke tasks.
* Updates docs.
* Bumps faabric dependency.
* Bumps code version.

Re. the race condition, we now update the flag `namespacesInitialised` after they are actually initialised. Otherwise, another worker could see that the namespaces have been initialised, yet see the vector empty, and thus throw an exception.

Re. the new env variable. Network namespaces must be created beforehand. In remote deployments, this happens in the worker's entrypoint. In local deployments, one can create and check network namespaces using the CLI.


src/system/NetworkNamespace.cpp
```diff
@@ -43,6 +41,8 @@ std::shared_ptr<NetworkNamespace> claimNetworkNamespace()
                 namespaces.emplace_back(
                   std::make_shared<NetworkNamespace>(netnsName));
             }
+
+            namespacesInitialised = true;
```
Shillaker, line 45: Interesting, I've not look at this code for a long time, but could we just have a single `UniqueLock` that covers the whole function? There shouldn't be too much contention on this function, so should be ok for performance.

src/system/NetworkNamespace.cpp
```diff
@@ -27,29 +27,23 @@ void returnNetworkNamespace(std::shared_ptr<NetworkNamespace> ns)
...
+    faabric::util::UniqueLock lock(namespacesLock);
     if (namespaces.empty() && !namespacesInitialised) {
-        faabric::util::UniqueLock lock(namespacesLock);
-
-        if (namespaces.empty() && !namespacesInitialised) {
-            namespacesInitialised = true;
-
-            // Note that the availability of namespaces depends on the Faasm
-            // configuration for the relevant host
-            // TODO - connect these two up properly
```
csegarragonz, line 38: afaict we are already taking the host config into account in the `NetworkNamespace` constructor.

Shillaker: 👍

Shillaker: The host config is used in the `NetworkNamespace` constructor, but AFAICT it's just to get the name. What this TODO means, is that the number of network namespaces that are actually created on the host is not in any way connected with the `nNamespaces` variable we have here (which is arbitrarily set to `cores * 10`). This means that this class could allocate a namespace which doesn't actually exist on the host. 

So the TODO should stay, but perhaps worded more explicitly.

Shillaker (changes requested):

src/system/NetworkNamespace.cpp
csegarragonz: :ok_hand:

kubasz:
If I can chip in a small comment, it'd be good for network namespaces to be completely no-op when disabled, just like the other isolation features are. I ran into NS number limits while having them disabled, because they are still being allocated even if disabled. I did a hacky workaround here, disabling them entirely would be a just a couple more changes https://github.com/auto-ndp/faasm/commit/55e05a74810b072ae7ef7e5d23199be83a7aa19d

.github/workflows/tests.yml
```diff
@@ -179,7 +179,7 @@ jobs:
         run: inv -r faasmcli/faasmcli python.clear-runtime-pyc
       # --- Test run ---
       - name: "Run the tests"
-        run: /build/faasm/bin/tests
+        run: /build/faasm/bin/tests --abort
```
csegarragonz, line 182: This makes the tests exit at the first failure. Otherwise it is sometimes hard to see which test has failed.

src/system/NetworkNamespace.cpp
```diff
@@ -21,36 +21,49 @@ std::mutex namespacesLock;
...
 {
+    faabric::util::UniqueLock lock(namespacesLock);
     if (namespaces.empty() && !namespacesInitialised) {
-        faabric::util::UniqueLock lock(namespacesLock);
+        conf::FaasmConfig& conf = conf::getFaasmConfig();
+        int nNamespaces = conf.maxNetNs;
+
+        // If network namespaces are disabled, overwrite the env. variable and
+        // create a single pointer to use as a placeholder
+        if (conf.netNsMode == "off") {
```
Shillaker, line 40: I can't immediately tell the difference between these two lines:

```
if (ns->getMode() == NetworkIsolationMode::ns_off) {
 ```
 
 and 
 
 ```
if (conf.netNsMode == "off") {
 ```
 
 Is there any? Seems like only one form  of this check should be used.

File: tests/test/system/test_network.cpp [hunk 5]
```diff
@@ -42,4 +42,31 @@ TEST_CASE("Test basic network properties", "[faaslet]")
...
     conf.reset();
 }
+
+TEST_CASE("Test running out of namespaces", "[faaslet][network]")
+{
+    conf::FaasmConfig& conf = conf::getFaasmConfig();
+    std::string originalNetNsMode = conf.netNsMode;
+
+    faabric::util::setEnvVar("NETNS_MODE", "on");
+    conf.reset();
```
Shillaker, line 52: The easiest way to do this sort of config change is:

```
TEST_CASE_METHOD(FaasmConfTestFixture, "Test running out of namespaces", "[faaslet][network]") {
    conf.netNsMode = "on";

    // Do some testing
}
```

Then you don't need to call `getFaasmConfig()`, or handle environment variables or keeping track of the original value.

Shillaker (changes requested):

Shillaker (approved):

PR #484, opened 2021-09-20
docker-compose environment merge

Shillaker:
We previously had four different docker-compose environments:

- `docker-compose.yml` - local cluster, used for quick-start and local cluster development.
- `docker-compose-dev.yml` - used for local development.
- `dist-test/docker-compose.yml` - used for running distributed tests.
- `docker/docker-compose-ci.yml` - used for replicating GHA environment (never used).

All were out of sync, and the first three were doing _almost_ the same thing. 

To reduce complexity, make it easier to reason about, and reduce the number of different containers floating around, I've merged them all into one single `docker-compose.yml` file. 

The only real differences between the use-cases was whether we mounted the local versions of machine code and wasm files. This is now controlled via the `FAASM_LOCAL_MOUNT` and `FAASM_BUILD_MOUNT` environment variables. 

Changes:

- Always use a single `docker-compose.yml` and `bin/cli.sh` script.
- Avoid hard-coding IPs related to dist tests, instead looking up by hostname.
- Move the remaining `dist-tests` scripts into `deploy/dist-test` to avoid having another top-level dir.
- Put scripts related to local cluster dev in `deploy/local`, and distributed tests in `deploy/dist-test`. This was to avoid further bloating the `bin` directory.
- Update docs to reflect changes.
- Update GHA build.


bin/cli.sh
```diff
@@ -11,63 +11,50 @@ export FAASM_VERSION=$(cat VERSION)
...
-    export FAASM_CLI_IMAGE=faasm/cli:${FAASM_VERSION}
-fi
-
-if [[ -z "$CPP_CLI_IMAGE" ]]; then
-    export CPP_CLI_IMAGE=faasm/cpp-sysroot:${CPP_VERSION}
-fi
-
-if [[ -z "$PYTHON_CLI_IMAGE" ]]; then
-    export PYTHON_CLI_IMAGE=faasm/cpython:${PYTHON_VERSION}
-fi
```
Shillaker, line 28: These defaults are already set in `.env`

docker-compose.yml
```diff
@@ -34,10 +34,11 @@ services:
       - minio
     volumes:
       - ./dev/faasm/build/:${FAASM_BUILD_MOUNT}
+      - ./dev/faasm-local/wasm/:/usr/local/faasm/wasm
```
Shillaker, line 37: Not having both wasm and machine code mounted made it very confusing when they were out of sync.

docs/development.md
```diff
@@ -199,23 +199,6 @@ which func_runner
...
-[`docker/docker-compose-ci.yml`](../docker/docker-compose-ci.yml) exactly match
-those in [`.github/workflows/tests.yml`](../.github/workflows/tests.yml).
-
-Then, run the container and work through the steps in the Github Actions file,
-to see where things might have gone wrong.
-
-```bash
-# To start the container
-docker-compose -f docker/docker-compose-ci.yml --env-file ./.env run cli /bin/bash
-```
```
Shillaker, line 217: This is no longer relevant as the CI build is much more complex.

docker-compose.yml
```diff
@@ -77,46 +83,92 @@ services:
...
     depends_on:
-      - worker
+      - redis-state
+      - redis-queue
+      - minio
     environment:
       - UPLOAD_HOST=${UPLOAD_HOST:-upload}
       - INVOKE_HOST=${INVOKE_HOST:-nginx}
-      - LOG_LEVEL=info
+      - LOG_LEVEL=debug
```
Shillaker, line 138: This is a development container, so `debug` is the most commonly needed logging.

docker-compose.yml
```diff
@@ -77,46 +83,92 @@ services:
...
+  # Distributed tests server
+  dist-test-server:
+    image: ${FAASM_CLI_IMAGE}
+    working_dir: /build/faasm
+    depends_on:
+      - redis-queue
+      - redis-state
+      - minio
+    environment:
+      - LOG_LEVEL=debug
```
Shillaker, line 156: Although this is run in the CI build, so should be `info`, the dist tests are still relatively unstable, and having `debug` has proven very useful (plus there aren't very many tests actually running).

tests/test/faaslet/test_chaining.cpp
```diff
@@ -10,14 +10,14 @@ namespace tests {
...
-    execFuncWithPool(call);
+    execFuncWithPool(call, true, 5000);
 }
 
 TEST_CASE("Test named function chaining", "[faaslet]")
 {
     faabric::Message call =
       faabric::util::messageFactory("demo", "chain_named_a");
-    execFuncWithPool(call);
+    execFuncWithPool(call, true, 5000);
```
Shillaker, line 20: These tests were timing out sometimes, so I upped the timeout.

csegarragonz (approved):
LGTM

PR #486, opened 2021-09-28
ensure wamr and wamr+sgx hash files are stored differently

csegarragonz:


csegarragonz:
For the changes in this PR to work, we need the latest WAMR CMake build in #472, but we also need these changes there for the tests to run, so merging everything together.

PR #501, opened 2021-10-13
Interaction with SGX Enclave Re-Design

csegarragonz:
In this PR I change the way we interact with a running SGX enclave from Faasm. 

For the moment, we just create one enclave per Faasm runtime instance. All executors in the instance share the same enclave, and we don't reuse WAMR modules at all (not even in the same executor running the same function as part of different messages).

### Outside the enclave

This PR introduces the `WAMREnclave` object, a wrapper around the singleton enclave. Each `SGXWWAMRWasmModule` has a pointer to one static instance of the `WAMREnclave`. Access to this resource is write-protected by an runtime-wide lock. We must allow for multiple reads in order to support function chaining.

### Inside the enclave

We simplify the function code as much as possible. The most important change is that there is now a `ModuleStore` where loaded SGX modules are stored. Each module is identified by their `funcStr` including the message ID, so the same module may be loaded twice, if used in different messages (even if in the same executor). Modules are removed from the enclave upon executor's `reset()`, and if the module store fills up, execution fails.

As mentioned before, writes to the store are protected by the `WAMREnclave` lock, and we support concurrent reads.

I leave as future work to change the way process error codes inside the enclave.

### Changelog

* Introduce the `WAMREnclave` object, to be shared among all `SGXWAMRWasmModule` instances.
* Use the term `native symbols` for symbol definition/resolution in the WASM boundary, and the term ECalls/OCalls for symbol definition/resolution in the SGX boundary.
* Re-factor the `ECalls` to remove the `faasm_sgx_` prefix.
* Re-factor in-enclave code to use (more) modern CPP.


tests/test/sgx/test_sgx_module.cpp
```diff
@@ -0,0 +1,146 @@
...
+    if (th.joinable()) {
+        th.join();
+    }
+}
+
+TEST_CASE_METHOD(SgxModuleTestFixture,
+                 "Test two modules executing different functions repeatedly",
+                 "[sgx2]")
+{
+    int numExec = 10;
```
csegarragonz, line 106: This is the test we were discussing offline

tests/test/sgx/test_sgx_module.cpp
```diff
@@ -0,0 +1,146 @@
...
+    // Note that both modules operate on the same enclave
+    std::thread th([this, numExec] {
+        otherModule.doBindToFunction(otherMsg, false);
+
+        for (int i = 0; i < numExec; i++) {
+            otherModule.executeFunction(otherMsg);
+        }
+    });
+
+    REQUIRE_NOTHROW(module.doBindToFunction(msg, false));
```
Shillaker, line 117: I don't think we need so many `REQUIRE_NOTHROW`s, what does this add to the test? `REQUIRE_NOTHROW` could technically be applied to every non-asserted line in every test.

tests/test/sgx/test_sgx_module.cpp
Shillaker: Ok great, couple of comments inlined.

What's the `sgx2` flag?

tests/test/sgx/test_sgx_module.cpp
```diff
@@ -0,0 +1,146 @@
...
+    }
+
+    // Flush enclave module
+    faabric::scheduler::getScheduler().flushLocally();
+
+    // Run functions again changing the module they run on
+    // Note that calling execute on a new function implicitly calls bind
+    std::thread th2([this, numExec] {
+        for (int i = 0; i < numExec; i++) {
+            otherModule.executeFunction(msg);
```
Shillaker, line 134: This approach of reusing a module to execute another function isn't expected to work in Faasm. Usually we create a new module, bind to a function, then it's bound for life, and dies otherwise. I think there are checks in the WAVM modules to check that it can't rebind to a different function. This is to simplify reasoning around module state.

tests/test/sgx/test_sgx_module.cpp
```diff
@@ -0,0 +1,146 @@
...
+                 "[sgx]")
+{
+    // We can unbind when nothing is bound
+    REQUIRE_NOTHROW(module.reset(msg, ""));
+
+    // Try binding to a function
+    REQUIRE_NOTHROW(module.doBindToFunction(msg, false));
+
+    // Then unbind
+    REQUIRE_NOTHROW(module.reset(msg, ""));
```
Shillaker, line 45: Same point re. all of these `REQUIRE_NOTHROWS`, it's implicit that a test will fail if there's an exception. You can add a comment to say "This test is just checking that no exception is thrown".

tests/test/sgx/test_sgx_enclave.cpp
```diff
@@ -0,0 +1,49 @@
...
+class SgxTestFixture
+{
+  public:
+    SgxTestFixture() {}
+
+    ~SgxTestFixture() {}
+
+  protected:
+    std::shared_ptr<sgx::WAMREnclave> wamrEnclave = nullptr;
+};
```
Shillaker, line 16: Is this fixture used anywhere? Might be useful to add the appropriate cleanup shared by all SGX tests and put it in the shared fixtures header file.

csegarragonz:
Closing after offline discusion.

PR #527, opened 2021-10-28
Fix Quick Start

csegarragonz:
This PR fixes three problems with the filesystem interaction after flush:
1. `boost::filesystem::remove_all(dir)` woudl also remove the directory it was called on. Albeit not necessarily an error, I am pretty sure this was not intended. Instead I add a helper method to remove all contents inside a directory.
2. `boost::filesystem::create_directories(path)` was throwing an uncaught exception complaining about "no such path or directory" as a consequence of 1.
3. The consolidated `std::filesystem::path::append` behaves differently: if two absolute paths are appended the first one is ignored.

For both problems I add a check in the tests that would fail without this PR. Additionally, I bump the cpp dependency to fix another problem with the python task flushing the workers.

I also change the `boost::filesystem` calls to `std::filesystem` _only_ in the files I change, not globally. Happy to re-factor everything now or in a separate PR.

Lastly, bump the version code after the last commit to make sure the quick start test actually picks up the latest code changes. This does not completely fix the problem until we have real per-commit e2e testing.


src/storage/FileLoader.cpp
```diff
@@ -378,17 +413,26 @@ std::vector<uint8_t> FileLoader::loadSharedObjectWasm(const std::string& path)
...
+    std::filesystem::directory_entry f(realPath);
     const std::string directory = f.path().parent_path().string();
     const std::string fileName = f.path().filename().string();
 
     // Work out the final destination for the object file. This will be the
     // object path with the directory of the original file appended
-    boost::filesystem::path objPath(conf.objectFileDir);
-    objPath.append(directory);
+    std::filesystem::path objPath(conf.objectFileDir);
+    objPath += directory;
```
csegarragonz, line 423: `path.append` works slightly different in `std::filesystem`. Namely if appending two paths both of which start with a backslash, then the first one will be ignored, and the second one taken by an absolute path (unless otherwise stated).

See [this post for more details](https://stackoverflow.com/questions/48347642/differences-between-c14-stdexperimentalfilesystemv1-and-c17-stdfiles/48349966#48349966).

tests/test/codegen/test_machine_code_generator.cpp
```diff
@@ -219,6 +217,9 @@ TEST_CASE_METHOD(CodegenTestFixture,
     // Run the codegen
     gen.codegenForSharedObject(localSharedObjFile);
 
+    // Check the locally cached path matches the expected one
+    REQUIRE(loader.getSharedObjectObjectFile(localSharedObjFile) == objFile);
```
csegarragonz, line 221: With the new `path.append` behaviour in `std::filesystem` this check would have failed.

This is, before:
`path.append("/tmp/obj", "/usr/local/faasm/foo/bar.o"); // = "/usr/local/faasm/foo/bar.o")`
Now:
`"/tmp/obj" += "/usr/local/faasm/foo/bar.o"; // = "/tmp/obj/usr/local/faasm/foo/bar.o"`

tests/test/faaslet/test_flushing.cpp
```diff
@@ -58,22 +58,25 @@ TEST_CASE_METHOD(FlushingTestFixture,
...
     // Check that the shared file exists
     std::string syncSharedPath = "faasm://" + fileName;
     REQUIRE(storage::SharedFiles::syncSharedFile(syncSharedPath, "") == 0);
 
     // Flush and check file is gone
     loader.clearLocalCache();
-    REQUIRE(!boost::filesystem::exists(sharedPath));
+
+    // Check the shared files dir exists, but not the specific shared file path
+    REQUIRE(std::filesystem::exists(conf.sharedFilesDir));
```
csegarragonz, line 73: Without the changes in `FileLoader::clearLocalCache()` this chech would fail.

Before, `remove_all(conf.sharedFilesDir)` would delete `conf.sharedFilesDir` as well.

Now, we use `removeAllInside` that recursively deletes the contents of `conf.sharedFilesDir` but not the directory itself.

src/storage/FileLoader.cpp
```diff
@@ -30,6 +28,30 @@ namespace storage {
 #define WAMR_AOT_FILENAME "function.aot"
 #define SGX_WAMR_AOT_FILENAME "function.aot.sgx"
 
+static std::uintmax_t removeAllInside(const std::filesystem::path& dir)
+{
+    std::uintmax_t removedItemsCount{ 0 };
+
+    if (not is_directory(dir)) {
```
Shillaker, line 35: We use `!`, not `not` in general. Would be good to stay consistent with the repo style.

Where is this `is_directory` coming from? Is it in `std::filesystem`? If so, we should prepend `std::filesystem` (as is done elsewhere in this file).

src/storage/FileLoader.cpp
Shillaker: OK sure, this seems like odd behaviour to me (and clearly to the `boost::filesystem` authors too), so could you put a comment above this as a warning? I also think you mean forward slash rather than backslash (unless you're running on windows 😄)

src/storage/FileLoader.cpp
```diff
@@ -57,17 +79,26 @@ static std::string getKey(const faabric::Message& msg,
...
+        try {
+            std::filesystem::create_directories(path);
+        } catch (const std::filesystem::filesystem_error& ex) {
+            SPDLOG_ERROR(
+              "Call to create_directories ({}) failed with error ({}): {}",
+              path.string(),
+              ex.code().value(),
+              ex.what());
+            throw std::runtime_error("Filesystem error creating directories");
+        }
```
Shillaker, line 101: This snippet of code calling `create_directories` is repeated three times. It should be in a function called something like `createDirectories` which can just be `static` in this file for now.

Shillaker (changes requested):
Couple of style changes but other than that looks good 👍

src/storage/FileLoader.cpp
```diff
@@ -30,6 +28,30 @@ namespace storage {
 #define WAMR_AOT_FILENAME "function.aot"
 #define SGX_WAMR_AOT_FILENAME "function.aot.sgx"
 
+static std::uintmax_t removeAllInside(const std::filesystem::path& dir)
+{
+    std::uintmax_t removedItemsCount{ 0 };
```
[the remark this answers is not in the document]
kubasz: `{}` has a different meaning (initialize without implicit conversions) vs `=` but it doesn't matter for 0

src/storage/FileLoader.cpp
kubasz: This is probably argument-dependent lookup (ADL) - dir's type is from std::filesystem:: so it gets included in the search path - I wonder if there's a clang-tidy lint to find these lookups automatically in case one slips in somewhere

Shillaker (approved):

PR #537, opened 2021-11-09
Use latest `ExecGraph` features and only upon request

csegarragonz:
In this PR I bump the faabric dependency to catch (1) the new exec graph details (faasm/faabric#166) and (2) the bug-fixes for the new exec graph details (faasm/faabric#173) :rofl: 

I also make recording execution graphs an opt-in feature that a user can enable with the `--graph` flag on any `inv invoke` call.

I bump the code version to make it available in production environments.


faasmcli/faasmcli/util/call.py
```diff
@@ -151,10 +155,10 @@ def status_call_impl(user, func, call_id, host, port, quiet=False):
         return STATUS_RUNNING, call_result
 
 
-def exec_graph_call_impl(user, func, call_id, host, port, quiet=False):
+def exec_graph_call_impl(call_id, host, port, quiet=False):
     msg = {
-        "user": user,
-        "function": func,
+        "user": "",
```
csegarragonz, line 160: Passing `None` here broke the JSON parser in faabric.

src/wasm/chaining_util.cpp
```diff
@@ -65,6 +65,10 @@ int makeChainedCall(const std::string& functionName,
         msg.set_issgx(true);
     }
 
+    if (originalCall->recordexecgraph()) {
+        msg.set_recordexecgraph(true);
+    }
+
```
Shillaker, line 71: Need tests for both setting this to `true` in chained messages (should be adding a line to an existing test), and the call to `logChainedFunction` only when this is on.

As a general rule of thumb, if you're fixing a bug, you should be writing a test that would have caught it in the first place. Although this feature is still somewhat experimental, it's still very worth testing.

Shillaker (changes requested):

tests/test/system/test_exec_graph.cpp
```diff
@@ -0,0 +1,90 @@
+#include <catch2/catch.hpp>
+
+#include "faasm_fixtures.h"
+
+#include <faabric/runner/FaabricMain.h>
+#include <faabric/scheduler/ExecGraph.h>
+#include <faabric/scheduler/Scheduler.h>
+// DELETE MEEEE
```
Shillaker, line 8: Do we need to do what it says?

tests/test/system/test_exec_graph.cpp
```diff
@@ -0,0 +1,90 @@
...
+#include <faabric/util/logging.h>
+#include <faaslet/Faaslet.h>
+
+namespace tests {
+class ExecGraphTestFixture : public FunctionExecTestFixture
```
Shillaker, line 13: We already have tests for the execution graph in `tests/test/scheduler/test_exec_graph.cpp`, I think these can just be put in there can't they? The exec graph doesn't really have any thing to do with the `system` module.

Shillaker (changes requested):

tests/test/system/test_exec_graph.cpp
csegarragonz: I completely missed this, sorry.

tests/test/system/test_exec_graph.cpp
csegarragonz: Those tests live in `faabric`, and in `faabric` we [already check](https://github.com/faasm/faabric/blob/f7a910bcb262777ca22b47646b7c05497bc63f44/tests/test/scheduler/test_mpi_exec_graph.cpp#L46) that if the flag is not set, the execution graph is not generated.

I thought your comment in the previous review meant that you wanted to have an e2e check in faasm.

tests/test/system/test_exec_graph.cpp
```diff
@@ -0,0 +1,88 @@
...
+{
+  public:
+    ExecGraphTestFixture()
+      : m(std::make_shared<faaslet::FaasletFactory>())
+    {
+        // Modify system config (network ns requires root)
```
Shillaker, line 17: I'm not sure we need this fixture, can we reuse one we already have (e.g. `FunctionExecTestFixture`)?

The `callFunctionGetResult` method seems unnecessary as the result isn't used AFAICT. Can we just use `execFunction`?

This comment also doesn't seem relevant, is it a copy-paste error?

tests/test/system/test_exec_graph.cpp
Shillaker: Gah sorry I'm getting confused now 😵, you're right, difficult to merge these tests with those in a different repo 😂 .

However, I still think we can find a better home for them. I think the first generic one can go into `tests/test/faaslet/test_chaining.cpp`, then the MPI one into `tests/test/faaslet/test_mpi.cpp`.

Shillaker (changes requested):

tests/test/system/test_exec_graph.cpp
csegarragonz: You are right, the fixture was unnecessary. The call to `sch.callFunction` is necessary to update the exec graph through faabric (thus we can't use `execFunction` which invokes the function directly on the module).

Shillaker (approved):

PR #538, opened 2021-11-12
Add distributed tests for MPI

csegarragonz:


tests/dist/mpi/test_remote_execution.cpp
```diff
@@ -0,0 +1,34 @@
...
+
+    // Set up the message
+    std::shared_ptr<faabric::BatchExecuteRequest> req =
+      faabric::util::batchExecFactory("mpi", "mpi_bcast", 1);
+    faabric::Message& msg = req->mutable_messages()->at(0);
+    msg.set_ismpi(true);
+    msg.set_mpiworldsize(4);
+
+    // Call the functions
+    sch.callFunctions(req);
```
Shillaker, line 27: It would be nice to somehow test that these had been executed on different hosts. Could we perhaps check the execution graph? Would be a nice indirect test of that working in a distributed environment too.

tests/dist/mpi/test_multi_tenant.cpp
```diff
@@ -0,0 +1,49 @@
...
+
+#include "fixtures.h"
+
+#include <faabric/scheduler/Scheduler.h>
+
+namespace tests {
+
+/* 12/11/2021 - This test deliberately fails as multi-tenant MPI execution is
+ * still not supported. We keep it here to showcase this is the use case we
+ * want to support.
```
Shillaker, line 11: Rather than commenting out tests like this, we should use the catch2 `[.]` tag which means they aren't run by default. That way they will still get compiled.

deploy/dist-test/upload.sh
```diff
@@ -24,6 +24,8 @@ docker-compose \
     cpp \
     inv func.user demo \
     func.upload-user demo \
+    func mpi mpi_bcast \
+    func.upload mpi mpi_bcast \
```
Shillaker, line 28: If it's not too inefficient it's probably easier just to do the whole `mpi` user as it would be nice to add more distributed MPI tests in future.

Shillaker (changes requested):

tests/dist/mpi/test_remote_execution.cpp
csegarragonz: Yes, I think this is a very valid point. I add a utility function in faasm/faabric#178 and then check for it here.

docker/cpp-root.dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM faasm/faabric-base:0.2.0
```
csegarragonz, line 1: This was needed to merge the latest conan changes in faabric.

Shillaker (approved):

PR #542, opened 2021-11-30
Cache conan dependencies in tests for faster build times

kubasz:
Just like in faabric, cache conan dependencies for faster test build times. Instead of a separate build step, just use the cache in the two containers directly as there's only 2, introducing a third one to pre-populate the cache would most likely by slower.


kubasz:
Looks like this halves the build time on a cache hit

Shillaker (approved):

PR #548, opened 2021-12-13
Snapshot and shared memory updates

Shillaker:
Update to new Faabric snapshots https://github.com/faasm/faabric/pull/176 https://github.com/faasm/faabric/pull/201

Support use of new CPP markup https://github.com/faasm/cpp/pull/71

Closes https://github.com/faasm/faasm/issues/545


bin/workon.sh
```diff
@@ -100,6 +100,15 @@ export PS1="(faasm) $PS1"
...
 ulimit -s 16384
 
+# -----------------------------
+# Sanitizers
+# -----------------------------
+# Check these match the CI build
+export ASAN_OPTIONS="halt_on_error=1"
+export LSAN_OPTIONS="suppressions=${PROJ_ROOT}/leak-sanitizer-ignorelist.txt"
+export TSAN_OPTIONS="halt_on_error=1:suppressions=${PROJ_ROOT}/thread-sanitizer-ignorelist.txt:history_size=7:second_deadlock_stack=1"
+export UBSAN_OPTIONS="print_stacktrace=1:halt_on_error=1"
```
Shillaker, line 110: This was previously listed in the dev set-up instructions where it will inevitably rot and must be copy-pasted regularly. Here it's set up by default. Didn't think this was worth a separate PR but can change.

src/storage/FileDescriptor.cpp
```diff
@@ -586,7 +585,6 @@ Stat FileDescriptor::stat(const std::string& relativePath)
         statResult.wasiFiletype = __WASI_FILETYPE_BLOCK_DEVICE;
     } else if (S_ISDIR(nativeStat.st_mode)) {
         statResult.wasiFiletype = __WASI_FILETYPE_DIRECTORY;
-        isReadOnly = true;
```
Shillaker, line 589: Unused `isReadOnly` caused compiler warnings.

include/wasm/WasmModule.h
```diff
@@ -169,20 +172,19 @@ class WasmModule
     int threadPoolSize = 0;
     std::vector<uint32_t> threadStacks;
 
-    std::shared_mutex moduleMemoryMutex;
-    std::mutex modulePthreadsMutex;
-    std::mutex moduleStateMutex;
```
Shillaker, line 174: It was quite hard to reason about these mutexes, so I tried to simplify with the `moduleMutex` (and one other object-specific mutex for the `sharedMemWasmPtrs`.

src/wasm/WasmModule.cpp
```diff
@@ -199,8 +252,31 @@ void WasmModule::restore(const std::string& snapshotKey)
...
     uint8_t* memoryBase = getMemoryBase();
     reg.mapSnapshot(snapshotKey, memoryBase);
+}
+
+void WasmModule::ignoreThreadStacksInSnapshot(const std::string& snapKey)
+{
+    std::shared_ptr<faabric::util::SnapshotData> snap =
+      faabric::snapshot::getSnapshotRegistry().getSnapshot(snapKey);
+
+    uint32_t threadStackRegionStart =
```
csegarragonz, line 262: Could we maybe add a comment here on the stack structure, or point somewhere where this structure may be; justifying in the process why we add `1`.

src/wasm/WasmModule.cpp
```diff
@@ -485,74 +567,74 @@ uint32_t WasmModule::createMemoryGuardRegion(uint32_t wasmOffset)
 void WasmModule::queuePthreadCall(threads::PthreadCall call)
 {
     queuedPthreadCalls.emplace_back(call);
+    isQueuedPthreadCalls.store(true);
```
csegarragonz, line 570: Again, we may want to pass `std::memory_order_release` and `std::memory_order_acquire_release` for these atomic operations (`store` and `fetch_add`).

src/wasm/WasmModule.cpp
```diff
@@ -485,74 +567,74 @@ uint32_t WasmModule::createMemoryGuardRegion(uint32_t wasmOffset)
...
 
 int WasmModule::awaitPthreadCall(const faabric::Message* msg, int pthreadPtr)
 {
     assert(msg != nullptr);
 
-    if (!queuedPthreadCalls.empty()) {
-        faabric::util::UniqueLock lock(modulePthreadsMutex);
+    bool isQueued = true;
+    bool mustDequeue =
+      isQueuedPthreadCalls.compare_exchange_strong(isQueued, false);
```
csegarragonz, line 580: Same re. memory order.

src/wasm/WasmModule.cpp
```diff
@@ -562,139 +644,19 @@ int WasmModule::awaitPthreadCall(const faabric::Message* msg, int pthreadPtr)
...
-    }
-
-    // Create ordered list of offsets
-    std::vector<uint32_t> sortedOffsets(ompLevel->sharedVarOffsets.get(),
-                                        ompLevel->sharedVarOffsets.get() +
-                                          ompLevel->nSharedVarOffsets);
-
-    std::sort(sortedOffsets.begin(), sortedOffsets.end());
+    // Decrease the count and see if we're the last one
+    int prevValue = queuedPthreadCallsCount.fetch_sub(1);
```
csegarragonz, line 648: Set memory order

src/wavm/WAVMModuleCache.cpp
```diff
@@ -58,6 +59,27 @@ WAVMModuleCache::getCachedModule(faabric::Message& msg)
...
 
+std::string WAVMModuleCache::registerResetSnapshot(wasm::WasmModule& module,
+                                                   faabric::Message& msg)
+{
+    std::string snapKey = faabric::util::funcToString(msg, false) + "_reset";
+
+    faabric::snapshot::SnapshotRegistry& reg =
+      faabric::snapshot::getSnapshotRegistry();
+
+    if (!reg.snapshotExists(snapKey)) {
```
csegarragonz, line 70: Same point regarding the pattern and the atomic checks.

tests/test/storage/test_file_descriptor.cpp
```diff
@@ -300,7 +300,8 @@ void checkWasiDirentInBuffer(uint8_t* buffer, DirEnt e)
 {
     size_t wasiDirentSize = sizeof(__wasi_dirent_t);
     auto wasiDirent = faabric::util::unalignedRead<__wasi_dirent_t>(
-      reinterpret_cast<std::byte*>(buffer));
+      reinterpret_cast<std::uint8_t*>(buffer));
```
csegarragonz, line 303: Same question re. `std::byte`.

csegarragonz (changes requested):
LGTM. Just minor comments re.`std::byte` and atomic variables. Also a reminder to finish off and merge the faabric PR first.

include/wavm/WAVMWasmModule.h
```diff
@@ -24,8 +24,8 @@ std::vector<uint8_t> wavmCodegen(std::vector<uint8_t>& wasmBytes,
 template<class T>
 T unalignedWavmRead(WAVM::Runtime::Memory* memory, WAVM::Uptr offset)
 {
-    const std::byte* bytes =
```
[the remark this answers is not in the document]
Shillaker: The interface for these functions is all `uint8_t` to be consistent with the rest of the codebase, so here we don't have a choice. Because the two aren't compatible, we would need to do a wholesale switch otherwise we'd have casts scattered around the place. 

In general changes like this are probably best done wholesale to make the testing and reviewing process simpler. Issu on the topic: https://github.com/faasm/faabric/issues/199

src/faaslet/Faaslet.cpp
```diff
@@ -100,7 +96,10 @@ int32_t Faaslet::executeTask(int threadPoolIdx,
     // Lazily setup Faaslet isolation.
     // This has to be done within the same thread as the execution (hence we
     // leave it until just before execution).
-    if (!isIsolated) {
+    bool notIsolated = false;
+    bool shouldIsolate = isIsolated.compare_exchange_strong(notIsolated, true);
```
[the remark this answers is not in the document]
Shillaker: Ah yes, I had thought this race would happen when an executor is handling multiple threads from the same batch request (in which case `executeTask` is called repeatedly on the same executor), but actually the cgroup and namespace operations are per thread and do their own internal locking, so there's no need to do this in the first place. I'll add a note in the code to that effect.

src/runner/func_runner.cpp
```diff
@@ -121,10 +121,14 @@ int main(int argc, char* argv[])
     storage::initFaasmS3();
     faabric::transport::initGlobalMessageContext();
 
+    PROF_BEGIN
```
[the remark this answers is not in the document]
Shillaker: I see the argument against committing code to mark up specific operations, as these may not be useful to everyone doing profiling. I personally think it's useful to leave the profiling markup around operations that _risk_ being slow in future. If someone is profiling the application it's probably because it's slow, and having existing profiling annotations around risky code is like giving that person a nudge in future and saying "make sure you confirm that xyz is still performant".

This markup is different as it's the main wrapper code that will be useful to everyone profiling this executable. Without it, the profiling information won't be printed and I found it confusing to remember what I needed to add and where.

src/wasm/WasmModule.cpp
```diff
@@ -78,50 +76,87 @@ wasm::WasmEnvironment& WasmModule::getWasmEnvironment()
...
     std::string snapshotKey = getAppSnapshotKey(msg);
 
     faabric::snapshot::SnapshotRegistry& reg =
       faabric::snapshot::getSnapshotRegistry();
 
-    if (reg.snapshotExists(snapshotKey)) {
-        SPDLOG_TRACE(
-          "Snapshot already exists for app {} ({})", msg.appid(), snapshotKey);
-    } else {
+    bool exists = reg.snapshotExists(snapshotKey);
```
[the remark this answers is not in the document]
Shillaker: The boolean itself isn't shared between threads, it's just a local stack variable, so I'm not sure having an atomic would make a difference here. The `snapshotExists` method is guarded by a read lock and is doing a lookup on a map whose modifications are in turn guarded by a write lock, so I _think_ it's ok.

src/wasm/WasmModule.cpp
Shillaker: On second thoughts, this operation is now only single threaded so this synchronisation isn't needed.

csegarragonz (approved):

PR #553, opened 2021-12-19
Fixes to image build process

csegarragonz:
In this PR I fix a series of issues I have run into whilst trying to release a new docker image:
* Pin conan dependencies to a specific revision. Otherwise, the same package versions can fail in subsequent builds if the revision for the package version changes upstream (see faasm/faabric#200).
* Use only virtual environments to handle python dependencies. Due to an issue with [pip's editable installs](https://github.com/pypa/pip/issues/7953), our python builds stopped working. A workaround seemed to be enabling `site.ENABLE_USER_SITE` in the package's `setup.py` (i.e. in `faasmtools` and `faasmcli`) but this derived in other problems. In the end I have decided to use only virtual environments around (as we do in regular development environments). Note that, to prevent a clash between container and out-of-container builds in the `venv` directory, I make out-of-container builds use `venv-bm` as their virtual environment root.
* Fix linking issue in `aws-sdk-cpp`. This conan package seems very unstable and fragile, we revert to installing it as an `ExternalProject` in CMake. I add an interface library + alias to make the replacement transparent.

Closes https://github.com/faasm/faasm/issues/324


csegarragonz:
This PR should be ready to go. Unfortunately, I now can't build the containers as GNU's FTP server seems to be down, or unreachable.

.github/workflows/tests.yml
```diff
@@ -256,21 +260,42 @@ jobs:
...
       - name: "Build codegen - Codegen Func"
-        run: inv -r faasmcli/faasmcli dev.cc codegen_func
+        run: |
+          source venv/bin/activate
+          inv -r faasmcli/faasmcli dev.cc codegen_func
+        shell: bash
       - name: "Build codegen - Codegen Shared Obj"
-        run: inv -r faasmcli/faasmcli dev.cc codegen_shared_obj
+        run: |
+          source venv/bin/activate
```
Shillaker, line 274: Can we avoid this repetition with a [custom shell](https://docs.github.com/en/actions/learn-github-actions/workflow-syntax-for-github-actions#custom-shell) or by setting up an alias for `inv` at the start?

cmake/ExternalProjects.cmake
```diff
@@ -58,10 +54,42 @@ find_package(Catch2 REQUIRED)
 # https://github.com/aws/aws-sdk-cpp/blob/main/Docs/CMake_External_Project.md
 # but they don't specify how to link the libraries, which required adding an
 # extra couple of CMake targets.
-# We now use the Conan package for aws-sdk-cpp: https://conan.io/center/aws-sdk-cpp
-# To enable only S3, all default targets from the Conan recipe
-# have to be explicitly disabled above
-find_package(AWSSDK REQUIRED)
+# There are some AWS docs on using the cpp sdk as an external project:
```
Shillaker, line 57: Would be good to put a warning here with a date saying why we're not using Conan for aws stuff in case somebody gets tempted.

.github/workflows/tests.yml
csegarragonz: I set a default shell (that activates the virtual env. before running the command) only for the job that runs the tests.

Note that we can't set the shell as a default for all the workflow as not all jobs run in the same container (thus don't have the venv set, neither they need it).

Shillaker (approved):
Awesome, nice job sorting this all out 👍

PR #564, opened 2022-01-11
OpenMP performance experiments

Shillaker:


Shillaker:
Split up into several smaller PRs.

PR #565, opened 2022-01-12
Migration PoC

Shillaker:
In this PR we implement the logic for our migration PoC based on function chaining with adequate input parameters. We rely on faasm/faabric#214 to detect the migration opportunities, and we do the rest here.

I add a distributed test that exercises the proof-of-concept.


src/wavm/faasm.cpp
```diff
@@ -1,6 +1,9 @@
+#include "faabric/snapshot/SnapshotClient.h"
+#include "faabric/util/scheduling.h"
```
Shillaker, line 2: If you see this, could you change these to `<>`?

PR #576, opened 2022-02-11
Generate documentation

Shillaker:
Changes:

- Add Doxygen doc generation for C++ code
- Integrate with Sphinx using Breathe
- Add full docs generation to Github Actions build

Docs are published at https://faasm.readthedocs.io/en/latest/

Note that the RTD build runs on the RTD servers, so the wrapper used in the GHA build is just to make sure it doesn't error, it doesn't actually produce the final result.


.github/workflows/tests.yml
```diff
@@ -33,6 +33,28 @@ jobs:
...
+      - name: "Check out branch"
+        run: git checkout --force ci-branch
+      - name: "TEMP - install doxygen"
+        run: apt install -y doxygen
+      - name: "TEMP - uninstall dodgy ansible"
+        run: pip uninstall ansible
+      - name: "TEMP - install python deps"
+        run: pip install -r faasmcli/requirements.txt
+      - name: "Build docs"
+        run: inv docs
```
Shillaker, line 56: I'll remove these `TEMP` steps when we do the next release.

PR #579, opened 2022-02-11
SGX: Re-run tests again as part of GHA.

csegarragonz:
Fix SGX execution, broken after the removal of `checkSgxSetup()` in faabric.

Make sure we are running SGX tests as part of our GHA pipelines.

### DIFF
--- a/faasmcli/faasmcli/tasks/dev.py
+++ b/faasmcli/faasmcli/tasks/dev.py
@@ -25,7 +25,7 @@ def cmake(
     perf=False,
     prof=False,
     sanitiser="None",
-    nosgx=True,
+    nosgx=False,
 ):
     """
     Configures the CMake build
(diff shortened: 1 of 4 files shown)


faasmcli/faasmcli/tasks/dev.py
```diff
@@ -25,7 +25,7 @@ def cmake(
     perf=False,
     prof=False,
     sanitiser="None",
-    nosgx=True,
+    nosgx=False,
```
csegarragonz, line 28: If we set this by default to `True`, I have found no way to re-enable SGX from the CLI.

I think this is because, as a `Boolean` value, it is interpreted as a flag. You can't set a Flag to `False` (i.e. not-Flag). Either you set it, or you don't. Therefore, if the flag is `--disable-sgx` and is `True` by default, there is no way to `--enable-sgx`.

I suggest we set it to `False` by default, and then if we don't want SGX we just run `--nosgx`. Further, we need to build with SGX by default for the tests to run.

Shillaker: Yes, sorry I think I meant to set it to `--nosgx=False` by default anyway. As you say, these flags have to have a default of false.

Shillaker (approved):

PR #580, opened 2022-02-11
Version 0.8.0

Shillaker:
- Master build fixes
- Fix task to build docs in GHA
- Disabled broken OpenMP tests (will fix in a later PR)
- Latest Faabric
- Fix Python environment in GHA
- Revert custom shell string introduced in #553 and use the more verbose approach (can simplify in a separate PR), as otherwise the tests just don't run at all.
- Use [`faasm/conan-cache-action`](https://github.com/faasm/conan-cache-action)


faasmcli/faasmcli/tasks/__init__.py
```diff
@@ -23,6 +23,7 @@
 
 # Default names
 ns = Collection(
+    bare_metal,
```
csegarragonz, line 26: @Shillaker I guess we are missing the code for this task?

faasmcli/requirements.txt
```diff
@@ -12,7 +12,7 @@ myst-parser==0.16.1
 Jinja2==2.11.3
 matplotlib==3.3.4
 networkx==2.5.1
-numpy==1.21.0
```
csegarragonz, line 15: @Shillaker is there a reason to unpin the version?

csegarragonz (approved):

faasmcli/faasmcli/tasks/__init__.py
Shillaker: What do you mean? The file is [here](https://github.com/faasm/faasm/blob/master/faasmcli/faasmcli/tasks/bare_metal.py). All the bare metal stuff is legacy and needs to be removed fully in a separate PR.

csegarragonz: Sorry, my point was that I did not understand why this file was modified (i.e. the `bare_metal` namespace added).

PR #581, opened 2022-02-11
Ensure tests actually run

csegarragonz:
In #553 we introduced a custom shell string to run a bash command, but activate the virtual environment beforehand. This was silently no-opping, and as a result the tests had not actually run in a couple of months.

In this PR I revert to a more verbose GHA syntax that certainly work, and fix the errors that had been upstreamed in the interim.


csegarragonz:
Closed in favour of #580

PR #583, opened 2022-02-14
Support for running SGX functions in `microbench_runner`

csegarragonz:
Add a flag in the `microbench_runner` to measure functions executed inside SGX. For vanilla WAMR we can set the environment variables.

### DIFF
--- a/docs/source/profiling.md
+++ b/docs/source/profiling.md
@@ -43,6 +43,12 @@ demo,hello,0,29.0,0.18
 These can then be parsed and plotted, as is done in the
 [experiment-microbench](https://github.com/faasm/experiment-microbench) repo.
 
+To run the same micro-benchmark in SGX, re-run the same command as follows:
+
+```bash
+WASM_VM="wamr" microbenchmark_runner <spec_file> <out_file> --sgx
+```
+
 ## High-level
 
 To get a quick overview of how things are performing you can use
--- a/src/runner/microbench_runner.cpp
+++ b/src/runner/microbench_runner.cpp
@@ -26,8 +26,13 @@ int main(int argc, char* argv[])
     initLogging();
 
     if (argc < 3) {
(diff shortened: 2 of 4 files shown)


docs/source/profiling.md
```diff
@@ -43,6 +43,12 @@ demo,hello,0,29.0,0.18
 These can then be parsed and plotted, as is done in the
 [experiment-microbench](https://github.com/faasm/experiment-microbench) repo.
 
+To run the same micro-benchmark in SGX, re-run the same command as follows:
+
+```bash
+WASM_VM="wamr" microbenchmark_runner <spec_file> <out_file> --sgx
```
Shillaker, line 49: Re. our offline conversation about making SGX a VM type, can we do that and change this to:

```bash
WASM_VM="sgx" microbenchmark_runner <spec_file> <out_file>
```

src/runner/microbench_runner.cpp
```diff
@@ -26,10 +26,15 @@ int main(int argc, char* argv[])
     initLogging();
 
     if (argc < 3) {
-        SPDLOG_ERROR("Usage: microbench_runner <infile> <outfile>");
+        SPDLOG_ERROR("Usage: microbench_runner <infile> <outfile> [--sgx]");
```
Shillaker, line 29: Doing `--` arguments like this is misleading as it's just a string and not actually an order-insensitive flag. There are a few ways to achieve the same effect: we could use an environment variable as mentioned above,
we could remove the `--` and use the third optional string for more general flags, we could add a `mode` or list of flags to the `microbench_runner` file format, or we could do proper commandline args with a library like Boost [program options](https://www.boost.org/doc/libs/1_78_0/doc/html/program_options.html).

In this case I think the `sgx` VM is the cleanest, and if that's not possible to that, a list of colon-separated flags could be passed in the file format, e.g.

```
demo,echo,10,wamr:sgx
```

Shillaker (changes requested):

PR #587, opened 2022-02-15
Set SGX execution through `WASM_VM` env. variable 

csegarragonz:
In this PR I change the way we indicate that we want to run a function with SGX. Even though it is not semantically correct, the environment variable indicating what `WASM_VM` we use takes now three possible values: `wavm` (default), `wamr`, and `sgx`.

This means that wen using the CLI we don't need to pass any extra parameters, just set the variable beforehand:

```bash
export WASM_VM="sgx"

# Codegen and run function inside enclave
inv codegen demo hello
inv run demo hello

export WASM_VM="wavm"

# Codegen and run function with WAVM
inv codegen demo hello
inv run demo hello
```

This also means that we can get rid of the references to SGX in the faabric message (faasm/faabric#229).

### DIFF
--- a/faasmcli/faasmcli/tasks/codegen.py
+++ b/faasmcli/faasmcli/tasks/codegen.py
@@ -4,7 +4,6 @@ from invoke import task
 from copy import copy
 from os import environ
 from os.path import join
-from sys import exit
 
 from faasmcli.util.codegen import find_codegen_func, find_codegen_shared_lib
 from faasmcli.util.env import FAASM_RUNTIME_ROOT
--- a/faasmcli/faasmcli/tasks/codegen.py
+++ b/faasmcli/faasmcli/tasks/codegen.py
@@ -44,25 +44,20 @@ SGX_WHITELISTED_FUNCS = [
 
 
 @task(default=True)
-def codegen(ctx, user, function, wamr=False, sgx=False):
+def codegen(ctx, user, function):
     """
     Generates machine code for the given function
     """
     env = copy(environ)
     env.update(
         {
-            "WASM_VM": "wamr" if wamr else "wavm",
             "LD_LIBRARY_PATH": "/usr/local/lib/",
         }
     )
 
-    if (not wamr) and sgx:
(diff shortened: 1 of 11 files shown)


faasmcli/faasmcli/tasks/codegen.py
```diff
@@ -14,7 +13,7 @@
     join(FAASM_RUNTIME_ROOT, "lib", "fake", "libfakeLibB.so"),
 ]
 
-WAMR_WHITELISTED_FUNCS = [
+WAMR_ALLOWED_FUNCS = [
```
csegarragonz, line 16: Kind of unrelated, but following our `master->main` re-factor, I thought I would also change this whilst I was amending this file (I don't think we use whitelist/blacklist anywhere else other than sgx code which will get re-factored now anyway).

Shillaker: Nice, yeah let's keep chipping away at this.

tests/test/codegen/test_machine_code_generator.cpp
```diff
@@ -112,20 +112,15 @@ TEST_CASE_METHOD(CodegenTestFixture,
...
-        SECTION("SGX")
-        {
-            objectFileA = "/tmp/obj/demo/hello/function.aot.sgx";
-            objectFileB = "/tmp/obj/demo/echo/function.aot.sgx";
-            msgA.set_issgx(true);
-            msgB.set_issgx(true);
-        }
+    SECTION("SGX codegen")
+    {
+        conf.wasmVm = "sgx";
```
Shillaker, line 121: Can you double check that there's a `FaasmConfigTestFixture` somewhere in the class hierarchy for this test? We just want to make sure that the config is reset at the end of each test.

Shillaker (approved):
Nice, much cleaner. Just a small comment about checking a test.

tests/test/codegen/test_machine_code_generator.cpp
csegarragonz: There is an `S3TestFixture` that resets the config :+1:

PR #593, opened 2022-02-17
Clean up enclave-related code running outside the enclave

csegarragonz:
In this PR I re-factor and do some clean-up of the code running outside the enclave (`{src,include}/enclave/outside`).

Changelog:
* Re-factor `SGXWamrWasmModule.{cpp,h}` to`EnclaveInterface.{cpp,h}`: as discussed offline, the real WASM module is inside the enclave, so this name was misleading.
* Re-factor `SGXWAMRWasmModule.S` to `getSgxSupport.S`: the name for the `ASM` script was also very misleading. This file checks for HW support for SGX. There are [other ways to do it](https://www.intel.com/content/www/us/en/developer/articles/technical/properly-detecting-intel-software-guard-extensions-in-your-applications.html) (not necessarily less involved), but in any case matter for a different PR.
* Factor-out the set of defined ECalls from `EnclaveInterface.h` to `ecalls.h` for clarity. Note that ECalls are declared outside the enclave, but defined inside*.
* Add method to process errors when doing ECalls. This was verbose and repeated before as we need to check for two error codes (one for SGX-system errors, one for Faasm-logic errors inside the enclave).

*This is not entirely true, as SGX generates stubbed calls using the `EDGER8R` tool, so a sort of ECall is defined outside the enclave as part of the build process. But this is not relevant to the discussion.


include/enclave/outside/ecalls.h
```diff
@@ -46,34 +38,3 @@ extern "C"
       sgx_enclave_id_t enclave_id,
       faasm_sgx_status_t* ret_val);
 }
-
```
csegarragonz, line 49: Weird diffing by Github here, but all this removed chunk here has been copied to `EnclaveInterface.h`

PR #594, opened 2022-02-17
Introduce `EnclaveWasmModule` to handle WASM modules inside enclaves

csegarragonz:


include/enclave/outside/ecalls.h
```diff
@@ -22,17 +19,17 @@ extern "C"
...
       sgx_enclave_id_t enclave_id,
       faasm_sgx_status_t* ret_val,
-      const unsigned int thread_id);
+      uint32_t faaslet_id);
 
     extern sgx_status_t faasm_sgx_enclave_call_function(
       sgx_enclave_id_t enclave_id,
       faasm_sgx_status_t* ret_val,
-      const uint32_t thread_id);
+      uint32_t faaslet_id);
```
Shillaker, line 32: Eventually it would be good to change to pascal case across all the SGX-related code, but that can be done in its own PR.

Shillaker (approved):

include/enclave/outside/ecalls.h
csegarragonz: Indeed, next PR is a re-factor of ECalls to pascal case.

PR #596, opened 2022-02-21
Fix `FAASM_SGX` flag

csegarragonz:
In a previous PR (#590) I had to overwrite the `FAASM_SGX` CMake variable to set it to `ON` explicitly to make sure the SGX code was being built and tested. My fix was too brute, and broke the option to disable the SGX build, which is why the flag was introduced in the first place.

The crux of the problem is that, even though the default value is now `ON`, it is _cached_ (in `CMakeCache.txt`) as `OFF` from when the container image was built. As a consequence, if we don't set `FAASM_SGX` to `ON` explicitely, it will be set to `OFF`. As a proof of concept, I print all cached variables in an action run, where [we can see that `FAASM_SGX` is set to OFF](https://github.com/faasm/faasm/runs/5277343740?check_suite_focus=true#step:4:1617).

One possible workaround is to re-build the container images with a different default value, so that it gets properly cached. Another workaround (the one I have chosen) is to `unset` the variable from the CMake cache so that it is either the default (`ON`) or `OFF` if overriden from the command line.

Relevant links:
- https://gitlab.kitware.com/cmake/cmake/-/issues/21708
- https://stackoverflow.com/questions/44719963/cmake-default-value-for-options-doesnt-work/44720601#44720601
- https://gitlab.kitware.com/cmake/cmake/-/issues/14756#note_258579

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -16,7 +16,6 @@ option(FAASM_PERF_PROFILING "Turn on profiling features as described in debuggin
 
 # SGX functionality
 # These options customise the SGX features _provided_ SGX is found
-unset(FAASM_SGX CACHE)
 option(FAASM_SGX "Enable SGX support" ON)
 option(FAASM_SGX_SIM_MODE "Turn on SGX sim mode" ON)
 option(FAASM_SGX_ATTESTATION "Turn on attestation" OFF)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -16,6 +16,7 @@ option(FAASM_PERF_PROFILING "Turn on profiling features as described in debuggin
 
 # SGX functionality
 # These options customise the SGX features _provided_ SGX is found
+unset(FAASM_SGX CACHE)
 option(FAASM_SGX "Enable SGX support" ON)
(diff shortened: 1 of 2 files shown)


csegarragonz:
@kubasz Could you check what I say makes any sense at all?

kubasz:
Because you're already using python scripts for everything, imo the best solution is to always pass `-DFAASM_SGX=value` to cmake, `on`/`off` and just not relying on the default value.

The cache keeps whatever was last used and really shouldn't be overwritten by the cmake scripts, as it's intended for user configuration and cmake will rerun itself with no extra arguments to regenerate build files keeping old cache settings.

Personally I've been using CMake presets (<https://cmake.org/cmake/help/v3.22/manual/cmake-presets.7.html>, <https://github.com/auto-ndp/faasm/blob/master/CMakePresets.json>) as they're very convenient for local builds and integrate well with vscode/clion unlike the python-based configuration, you may want to look into those to have a couple of pre-set configuration sets like docker, tests, local build, etc.

csegarragonz:
Ok thanks, I was not sure about tampering with the cache manually. To fix the current bug I think it makes more sense to just amend the CLI script.

Shillaker (approved):
Wow, good catch!

PR #606, opened 2022-03-03
Re-factor remaining SGX code to PascalCase

csegarragonz:
As far as I can tell, I haven't left anything behind.


Shillaker (approved):
Nice, can you scan through the rest of the SGX code and see if this is the last bit of non-Pascal code? Can bundle anything else into this PR.

PR #607, opened 2022-03-03
Run SGX locally in HW mode

csegarragonz:
In this PR I introduce the necessary changes to run the `demo/hello` function in an SGX enclave in hardware mode. Most importantly, we need to install Intel's PSW package alongside the SGX SDK. We build it as part of the SGX image, but install it in the `base` image, as some libraries need to be installed in `/usr/lib` (and would be lost in the multi-stage build).

I also unify the different configuration variables and boil it down to one, `FAASM_SGX_MODE` that can take as values: `"Disabled"`, `"Simulation"`, and `"Hardware"`. The default mode is simulation, but a different mode can be set using invoke tasks:

```bash
inv dev.cmake --sgx "Disabled"|"Simulation"|"Hardware"
```

For the moment, the only way to run functions in hardware mode is to run the client container in an SGX-enabled machine, and re-build the sources from scratch in "Hardware" mode. In a subsequent PR I will add support to run SGX in hardware mode in a k8s cluster.

I have tested myself that "Hardware" mode works in an SGX-enabled machine. Unfortunately, there is no direct way to check for this in the tests, unless we set up a Github runner in an Azure VM (which would be quite expensive).

Lastly, I update the method we use to check for SGX support, and add an invoke task to check if the machine the CLI is running on is SGX-capable and SGX-enabled, to use it:

```bash
inv dev.cc detect_sgx
inv sgx.check
```


cmake/FindSGX.cmake
```diff
@@ -1,331 +1,14 @@
-# Copied from
```
csegarragonz, line 1: This file defined a bunch of environment variables and CMake functions that we do not need/use. At this stage, we just want to check that the Intel SDK is installed under `/opt/intel/sgxsdk`. Any different installation will probably break our build anyway.

CMakeLists.txt
```diff
@@ -88,44 +87,42 @@ list(PREPEND CMAKE_PREFIX_PATH ${CMAKE_CURRENT_BINARY_DIR}/faabric)
...
-        # Must link protobuf statically because SGX LE probably uses an older version
-        set(Protobuf_USE_STATIC_LIBS 1)
-    endif()
-
-    if(FAASM_SGX_ATTESTATION)
-        add_definitions(-DFAASM_SGX_ATTESTATION=1)
-    else()
-        add_definitions(-DFAASM_SGX_ATTESTATION=0)
-    endif()
+    # SGX configuration
```
csegarragonz, line 99: We only try to find the SGX installation if SGX is not disabled from the config.

If we look for it, and don't find it, we must error out.

Note that this could only possibly happen in out-of-container builds.

cmake/ExternalProjects.cmake
```diff
@@ -122,18 +122,3 @@ message(STATUS FAASM_WAVM_SOURCE_DIR ${FAASM_WAVM_SOURCE_DIR})
 
 FetchContent_GetProperties(wamr_ext SOURCE_DIR WAMR_ROOT_DIR)
 message(STATUS WAMR_ROOT_DIR ${WAMR_ROOT_DIR})
-
-# SGX-specific dependencies
```
csegarragonz, line 126: This was long-ago removed, but never from here.

CMakeLists.txt
```diff
@@ -14,12 +14,11 @@ endif ()
...
-# These options customise the SGX features _provided_ SGX is found
-option(FAASM_SGX "Enable SGX support" ON)
-option(FAASM_SGX_SIM_MODE "Turn on SGX sim mode" ON)
-option(FAASM_SGX_ATTESTATION "Turn on attestation" OFF)
-option(FAASM_SGX_XRA "Turn on support for XRA" OFF)
+# This option customises the SGX features _provided_ SGX is found:
+# - Disabled: disable all SGX features
+# - Simulation: use SGX in simulation mode (default)
+# - Hardware: use SGX in hardware mode
+option(FAASM_SGX_MODE "Enable SGX support" "Simulation")
```
Shillaker, line 21: I think these terms are pretty self-explanatory but now the help text for the option isn't so clear, instead maybe:

```
option(FAASM_SGX_MODE "Type of SGX support: Disabled, Simulation or Hardware" "Simulation")
```

CMakeLists.txt
```diff
@@ -88,44 +87,36 @@ list(PREPEND CMAKE_PREFIX_PATH ${CMAKE_CURRENT_BINARY_DIR}/faabric)
...
 
-# Detect SGX
+# ----------------------------------------
+# SGX configuration
+# ----------------------------------------
+
 list(APPEND CMAKE_MODULE_PATH ${CMAKE_CURRENT_LIST_DIR}/cmake)
-if (FAASM_SGX)
+if (NOT (FAASM_SGX_MODE STREQUAL "Disabled"))
+    # Set maro definitions for conditional compilation
```
Shillaker, line 96: Typo

docker/base.dockerfile
```diff
@@ -2,7 +2,7 @@
 FROM faasm/cpython:0.1.0 as python
 
 # Import from SGX container
-FROM faasm/sgx:0.5.3 as sgx
+FROM faasm/sgx:0.8.1 as sgx
```
Shillaker, line 5: How come this isn't `0.8.2`? Looks like the `faasm/sgx` image has changed in this build.

docker/base.dockerfile
```diff
@@ -14,10 +14,13 @@ ENV FAASM_DOCKER="on"
...
-# Set up SGX SDK
+# Set up SGX SDK, and install the SDK and PSW libraries
 COPY --from=sgx /opt/intel /opt/intel
-RUN apt-get update
-RUN apt-get install -y nasm
+COPY --from=sgx /linux-sgx/linux/installer/bin/ /tmp/linux-sgx
+WORKDIR /opt/intel
+RUN sh -c 'echo yes | /tmp/linux-sgx/sgx_linux_x64_sdk_2.15.101.1.bin'
+RUN sh -c 'echo yes | /tmp/linux-sgx/sgx_linux_x64_psw_2.15.101.1.bin --no-start-aesm'
+RUN cp /opt/intel/sgxsdk/lib64/libsgx_capable.so /usr/lib
```
Shillaker, line 23: This looks like SGX setup that should be included in the `sgx` base image, why does it need to be here?

Also why do we need to copy `/opt/intel/sgxsdk/lib64/libsgx_capable.so` to `/usr/lib`? I would expect `FindSGX.cmake` to still exist, find this library where it's installed, then set this as part of an `SGX_LIBS` variable that things could link with.

docker/sgx.dockerfile
```diff
@@ -1,16 +1,46 @@
...
+    debhelper \
+    git \
+    libcurl4-openssl-dev \
+    libprotobuf-dev \
+    libssl-dev \
+    libtool \
+    lsb-release \
+    ocaml \
+    ocamlbuild \
+    protobuf-compiler \
```
Shillaker, line 22: Do we really need all these packages just to install SGX e.g. `protobuf-compiler` and OCaml? Seems like a lot.

docs/source/sgx.md
```diff
@@ -6,8 +6,26 @@ support using [WAMR](https://github.com/bytecodealliance/wasm-micro-runtime).
 
 ## Quick start
 
-SGX support is currently only available for development clusters.
-Note that the following will run in simulation mode.
+To configure SGX, we must build the code with the desired SGX flavour: disabled
+(no SGX functionality), simulation (default), or hardware.
+
+```bash
+inv dev.cmake --sgx "Disabled"|"Simulation"|"Hardware"
```
Shillaker, line 13: Do we need quotes around these values?

faasmcli/faasmcli/tasks/dev.py
```diff
@@ -65,8 +65,8 @@ def tools(ctx, clean=False, build="Debug", parallel=0, sanitiser="None"):
     """
     Builds all the targets commonly used for development
     """
-    nosgx = sanitiser != "None"
-    cmake(ctx, clean=clean, build=build, sanitiser=sanitiser, nosgx=nosgx)
+    sgx = "Disabled" if sanitiser != "None" else "Simulation"
+    cmake(ctx, clean=clean, build=build, sanitiser=sanitiser, sgx=sgx)
```
Shillaker, line 69: Several magic strings in this function now, would be nice to put them as constants at the top of the file, e.g.

```
SGX_MODE_SIM = "Simulation"
SGX_MODE_DISABLED = "Disabled"
SANITISER_NONE = "None"
```

src/enclave/outside/getSgxSupport.cpp
```diff
@@ -0,0 +1,69 @@
+#include <enclave/outside/getSgxSupport.h>
+#include <faabric/util/logging.h>
+
+#include <sgx_capable.h>
+
+namespace sgx {
+
+bool isSgxEnabled()
+{
+    sgx_status_t sgxStatus;
```
Shillaker, line 10: Can we define this inline where it's initialised?

```
sgx_status_t sgxStatus = sgx_is_capable(&isSgxCapable);
```

on line 14.

Shillaker (changes requested):
Nice one, this is much cleaner and seems a lot more robust. A few comments/ changes here and there.

docker/base.dockerfile
csegarragonz: The SGX image needs to be re-built manually with `inv docker.build -c sgx` and it picks up whatever version your working environment is in.

I bumped the code version last, what means I generated the image with the `0.8.1` version. I can re-generate it again with the new tag, but the code version makes no difference in the image build.

docker/base.dockerfile
csegarragonz: This is a bit involved, and I tried to clarify it in comments. I will answer questions one by one:

> This looks like SGX setup that should be included in the sgx base image, why does it need to be here?

The second installer (PSW) installs some libraries under `/usr/lib`. Currently, in `base.dockerfile` we inherit from `cpp-sysroot` and just _copy_ the `/opt/intel` directory from the sgx image, missing the libraries installed in `/usr/lib`, needed in [HW mode](https://github.com/intel/linux-sgx/issues/47).

> Also why do we need to copy `/opt/intel/sgxsdk/lib64/libsgx_capable.so` to `/usr/lib`?

We need to link against this library both in `"Hardware"` and `"Simulation"` builds, as it helps detecting the status of SGX. Unfortunately, it is not installed to `/usr/lib`, only to `/opt/intel/sgxsdk/lib64/libsgx_capable.so`. Note that the libraries installed under `/usr/lib` also live in `/opt/intel/sgxsdk/lib64`. As a consequence, if `target_link_libraries`-ing to `libsgx_capable.so` in the latter path, the duplicated libraries in the former path (`/usr/lib`) will be shadowed by the ones in the latter (`/opt/intel/sgxsdk/lib64`), yielding the same error as before.

CMakeLists.txt
```diff
@@ -88,44 +87,36 @@ list(PREPEND CMAKE_PREFIX_PATH ${CMAKE_CURRENT_BINARY_DIR}/faabric)
...
+    if (FAASM_SGX_MODE STREQUAL "Hardware")
+        add_definitions(-DFAASM_SGX_HARDWARE_MODE)
+    elseif (FAASM_SGX_MODE STREQUAL "Simulation")
+        add_definitions(-DFAASM_SGX_SIMULATION_MODE)
+    else ()
+        message(FATAL_ERROR "Unrecognised SGX mode: ${FAASM_SGX_MODE}")
+    endif ()
```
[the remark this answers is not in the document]
csegarragonz: AFAICT the variable has a default value, through the `option` command, to `Simulation`. I could change it to `Disabled` if you prefer, but this would prevent the SGX tests from running (by default).

docker/sgx.dockerfile
csegarragonz: Yes, we are building and installing the code for both the SDK and PSW. The dependency list is a combination of:
* The dependencies listed in the [Pre-requisites](https://github.com/intel/linux-sgx#prerequisites)
* The [docker image](https://github.com/intel/linux-sgx/blob/f607f89ab1ac648c3e62a23991befdeb93f9bf9d/docker/build/Dockerfile#L33-L50) they provide (note that they use `18:04` as base, what changes some packages).

CMakeLists.txt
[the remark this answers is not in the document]
Shillaker: Ah ok yes, sorry I was getting confused with this issue we saw before: https://github.com/faasm/faasm/pull/596. Will we see a similar problem here?

docker/base.dockerfile
Shillaker: Ah yes ok, because the image doesn't actually check out the code it doesn't really matter. Makes sense.

docker/base.dockerfile
Shillaker: Ok great, mentioning this in a comment in the code to explain would be good so that someone doesn't attempt to move it into the `sgx` container (which would have been my gut reaction).

docker/sgx.dockerfile
Shillaker: Ok wow, I didn't expect it to require so much stuff. Would be good to link those in a comment to explain why it's such a big list.

CMakeLists.txt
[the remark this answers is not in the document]
csegarragonz: I have followed Jakub's advice and always set the variable through the Python tasks, instead of relying on CMake's variable caching.

docker/sgx.dockerfile
csegarragonz: I add one comment to `sgx.dockerfile`.

Shillaker (approved):

PR #608, opened 2022-03-03
Move dev cluster scripts to Invoke tasks

Shillaker:
It was a bit clunky having the dev cluster stuff in a shell script, so I moved it all to some invoke tasks within the CLI.

Some of these functions only provider a very thin layer around the `docker-compose` CLI, but given that this is a common entrypoint for new/ unfamiliar users of the system, I think it's ok to put the most common operations in the same place.


csegarragonz (approved):
I agree this is very nice to have

PR #612, opened 2022-03-11
Run SGX in HW mode in k8s

csegarragonz:
In this PR I Introduce the necessary changes to allow running SGX functions in hardware mode in a kubernetes cluster. Most importantly, we need to build different image tags for workers with a hardware-mode SGX build.

In addition, when deploying the KNative services we need to: use the new worker docker image, and set the environment variable `WASM_VM=sgx` for both the upload and worker pods.
For the upload service, we can either use templated deployment files/env. var substitution or have separate deployment files. For the worker service I think we need a different file, as we have to add sgx-specific reources and we may have to add further options in the worker deployment file for attestation.

I have manually checked that with the current setup, and deploying as follows, we can run SGX functions in HW mode in AKS:

```bash
cd ~/experiment-base
inv cluster.provision --vm Standard_DC4s_v3 --nodes 4 --location eastus2 --sgx
inv cluster.credentials

cd ~/faasm
inv knative.install # run twice
inv knative.deploy --replicas=4 --sgx

inv upload demo hello <path_to_wasm>
inv invoke demo hello
```

I also bump the code version to test the new "Release" workflow file.


docker/base.dockerfile
```diff
@@ -51,12 +51,13 @@ RUN mkdir -p /usr/local/faasm/runtime_root/share
...
 
 # Build the basics here to set up the CMake build
+ARG FAASM_SGX_MODE
 RUN cmake \
     -GNinja \
     -DCMAKE_CXX_COMPILER=/usr/bin/clang++-13 \
     -DCMAKE_C_COMPILER=/usr/bin/clang-13 \
     -DCMAKE_BUILD_TYPE=Release \
-    -DFAASM_SGX_MODE=Simulation \
+    -DFAASM_SGX_MODE=${FAASM_SGX_MODE:-Simulation} \
```
csegarragonz, line 60: If `FAASM_SGX_MODE` is not set, use `Simulation`, [link to bash wiki](https://wiki.bash-hackers.org/syntax/pe#use_a_default_value).

docker/worker.dockerfile
```diff
@@ -1,5 +1,6 @@
 ARG FAASM_VERSION
-FROM faasm/base:${FAASM_VERSION}
+ARG FAASM_SGX_MODE
+FROM faasm/base${FAASM_SGX_MODE:+-sgx}:${FAASM_VERSION}
```
csegarragonz, line 3: If `FAASM_SGX_MODE` is not set, leave empty, if it is set, use `-sgx`. [Link to bash wiki.](https://wiki.bash-hackers.org/syntax/pe#use_an_alternate_value)

deploy/k8s/upload-sgx.yml
```diff
@@ -0,0 +1,43 @@
...
+      env:
+        - name: REDIS_STATE_HOST
+          value: redis-state
+        - name: REDIS_QUEUE_HOST
+          value: redis-queue
+        - name: LOG_LEVEL
+          value: "info"
+        - name: LD_LIBRARY_PATH
+          value: "/build/faasm/third-party/lib:/usr/local/lib"
+        - name: WASM_VM
```
csegarragonz, line 27: This file is a copy of `deploy/k8s/upload.yml` with just one extra environment variable.

deploy/k8s/worker-sgx.yml
```diff
@@ -0,0 +1,62 @@
...
+            - podAffinityTerm:
+                labelSelector:
+                  matchLabels:
+                    app: faasm-worker
+                topologyKey: kubernetes.io/hostname
+              weight: 100
+
+      containers:
+        - image: faasm/worker-sgx:0.8.3
+          resources:
```
csegarragonz, line 26: We need to add this `resources` field to make sure the SGX device is available in the pods ([see azure's docs](https://docs.microsoft.com/en-us/azure/confidential-computing/confidential-nodes-aks-overview)).

faasmcli/faasmcli/tasks/sgx.py
```diff
@@ -10,3 +15,35 @@ def check(ctx):
...
+            "--build-arg FAASM_SGX_MODE=Hardware",
+            "-f {} .".format(dockerfile),
+        ]
+        docker_cmd = " ".join(cmd)
+        print(docker_cmd)
+        run(docker_cmd, shell=True, check=True, cwd=PROJ_ROOT, env=shell_env)
+
+        docker_push_cmd = "docker push {}".format(tag_name)
+        print(docker_push_cmd)
+        run(docker_push_cmd, shell=True, check=True, cwd=PROJ_ROOT)
```
Shillaker, line 49: This looks like it's duplicating a lot of logic from the existing container build tasks (https://github.com/faasm/faasm/blob/main/faasmcli/faasmcli/tasks/docker_tasks.py), can we put it in there? It would just need to set the SGX mode conditionally based on the container name, e.g. `sgx_mode = "Hardware" if c.endswith("-sgx") else "Simulation"`, then strip the `-sgx` extension if it was present.

Shillaker (changes requested):

docker/base.dockerfile
csegarragonz: I always get confused with bash variable's expansion, so added the link and the explanation more for me than for you :rofl:

Shillaker (approved):
Nice, looks good. 

Can you triple check that running `inv version.bump` still works across all the k8s files? Looks like the task does whatever it finds in `deploy`, but just want to make sure.

deploy/k8s-sgx/upload.yml
```diff
@@ -0,0 +1,43 @@
...
+kind: Service
+metadata:
+  name: upload-lb
+  namespace: faasm
+spec:
+  type: LoadBalancer
+  ports:
+  - port: 8002
+  selector:
+    role: upload
```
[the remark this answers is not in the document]
csegarragonz: good catch, thanks. i copied the file before removing this :sweat:

PR #622, opened 2022-03-14
SGX-free builds

### LINKED ISSUE #617: Completely remove SGX from normal build and test
Now that we will have separate SGX containers, we can remove SGX from the default container, and have two test runs in GHA:

* Normal build and tests: runs the build in the default containers with `FAASM_SGX_MODE=Disabled`, and does all the normal sanitiser stuff
* SGX build and tests: runs the build in the SGX containers with `FAASM_SGX_MODE=Simulation`, then execute all the tests including the SGX ones (but none of the sanitiser stuff).

This is better than the situation we have now, as at the moment we don't check the build on a system without SGX installed (and this is the default expected setup for users).

csegarragonz:
In this PR I remove SGX completely from the default (`base`,`worker`,`cli`) containers. As suggested, I re-purpose the `sgx.dockerfile` into `base-sgx.dockerfile` that inherits from `base` with SGX installed. As a result, we generate the following new docker images: `base-sgx`, `base-sgx-sim`, `worker-sgx`, `worker-sgx-sim`, `cli-sgx`, `cli-sgx-sim`. All of them are created as a parallelised `Release` workflow file with the following stages and dependencies (images in the same column are built in parallel through a matrix):

```
            ┌─  worker
            │
minio    ┌──┼── cli                ┌─ worker-sgx
         │  │                      │
redis    │  └─  upload             ├──worker-sgx-sim
         │                         │
base ────┴──┬── base-sgx   ────┐   ├──cli-sgx
            │                  ├───┤
            └─  base-sgx-sim  ─┘   └─ cli-sgx-sim
```

Closes #617


.github/workflows/release.yml
```diff
@@ -16,9 +16,12 @@ jobs:
       - name: "Prune docker"
```
csegarragonz, line 16: This diff is quite confusing, you can browse the file [here](https://github.com/faasm/faasm/blob/sgxless/.github/workflows/release.yml).

.github/workflows/tests.yml
```diff
@@ -218,6 +218,98 @@ jobs:
       - name: "Run the tests"
         run: /build/faasm/bin/tests
 
+  sgx-tests:
```
csegarragonz, line 221: I don't merge this with `tests` as a single job with a matrix because `tests` already has the `sanitisers` matrix, and we only want to run the SGX tests with no sanitisers.

Shillaker (approved):
Awesome, actually a lot cleaner than I thought it would be :+1:

PR #623, opened 2022-03-16
Fixes for dev cluster environment

Shillaker:
- We weren't actually using the `env` set up in the task.
- Adding `FAASM_BUILD_DIR` in accordance with local dev setup.

### DIFF
--- a/faasmcli/faasmcli/tasks/cluster.py
+++ b/faasmcli/faasmcli/tasks/cluster.py
@@ -39,7 +39,7 @@ def start(ctx, workers=2, sgx=FAASM_SGX_MODE_DISABLED):
     ]
     cmd = " ".join(cmd)
     print(cmd)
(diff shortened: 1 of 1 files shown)


faasmcli/faasmcli/tasks/cluster.py
```diff
@@ -18,7 +19,10 @@ def start(ctx, workers=2, sgx=FAASM_SGX_MODE_DISABLED):
     """
     Start the local dev cluster
     """
+    # This env makes sure we mount our local setup into the containers, rather
+    # than using the prebuilt binaries
     env = copy(os.environ)
+    env["FAASM_BUILD_DIR"] = join(PROJ_ROOT, "dev/faasm/build")
```
Shillaker, line 25: `FAAASSM_BUILD_DIR` would get set by `bin/workon.sh` which would pick up the native build dir, not this one.

csegarragonz (approved):

PR #633, opened 2022-04-06
Bump CPP and Python versions

csegarragonz:
Both submodules currently point to dangling commits ([cpp](https://github.com/faasm/cpp/tree/1b1c2dceddf5d01e032a1395639fa4d7a029fd46), [python](https://github.com/faasm/python/tree/a8f1fac6ee3e8e4441bfeda5213d9b0fefe1d6bc)).


csegarragonz:
There's one WAMR test that keeps failing. There seems to be a clear error message, so I could take a look after I am done with the attestation PR.

Shillaker (approved):

PR #643, opened 2022-05-04
Changes for latest MPI

csegarragonz:
@Shillaker I am using version `0.8.11` in this PR to avoid clashes with #642


Shillaker (approved):
With faabric updated to `main` branch.

PR #645, opened 2022-05-09
Fixes to attest SGX enclaves in K8s pods

csegarragonz:
It is a requirement for K8s pods running code inside enclaves to have access to a volume mounted in `/var/run/aesmd.sock`. Unfortuntaely, we can not do that with Knative, as it supports a very small subset of mount options.

As a consequence, when deploying an SGX-enabled cluster, we have to get rid of Knative and use a standard `Deployment` (the same we use for native OpenMPI experiments).

A small observation is that, even though we don't use Knative under the hood, the task to deploy an SGX-enabled Faasm cluster still uses the `knative` task:

```bash
# To deploy
inv knative.deploy --replicas=4 --sgx

# To delete the workers
inv knative.delete-worker --sgx
```


deploy/k8s-sgx/worker.yml.j2
```diff
@@ -26,11 +33,17 @@ spec:
...
           # https://docs.microsoft.com/en-us/azure/confidential-computing/confidential-nodes-aks-overview#azure-device-plugin-for-intel-sgx-
+          name: faasm-worker
+          ports:
+            - containerPort: 8080
           resources:
             limits:
               sgx.intel.com/epc: "10Mi"
             requests:
               sgx.intel.com/epc: "10Mi"
+          volumeMounts:
```
csegarragonz, line 44: This is the keyword that is not supported in Knative.

faasmcli/faasmcli/tasks/knative.py
```diff
@@ -92,9 +106,33 @@ def deploy(ctx, replicas=0, sgx=False):
...
+    if sgx:
+        _deploy_faasm_worker_sgx(replicas)
+    else:
+        _deploy_faasm_worker(replicas)
 
-    ini_file(ctx)
+    ini_file(ctx, sgx=sgx)
+
+
+def wait_for_faasm_pods(label):
```
csegarragonz, line 117: Encapsulated this bit of code in its own method.

csegarragonz:
Closeing in beneffit of #661

PR #646, opened 2022-05-13
Lates faabric

csegarragonz:


docker/cpp-root.dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM faasm/faabric-base:0.3.0
+FROM faasm/faabric-base:0.3.3
```
csegarragonz, line 1: We need to bump this one for the latest Conan

tests/dist/fixtures.h
```diff
@@ -61,4 +63,67 @@ class DistTestsFixture
...
+      const faabric::scheduler::ExecGraph& execGraph,
+      const std::vector<std::string> expectedHosts)
+    {
+        std::vector<std::string> hostForRank =
+          faabric::scheduler::getMpiRankHostsFromExecGraph(execGraph);
+        REQUIRE(expectedHosts == hostForRank);
+    }
+
+    // TODO - this method is duplicated from faabric's dist tests
+    void checkSchedulingFromExecGraph(
```
csegarragonz, line 80: With this method we can actually check if a migration has taken place.

Shillaker (changes requested):
Also pending https://github.com/faasm/faabric/pull/264 as noted in the description.

tests/dist/mpi/test_migration.cpp
```diff
@@ -28,15 +27,80 @@ TEST_CASE_METHOD(DistTestsFixture, "Test migrating an MPI execution", "[mpi]")
...
+    // Try to migrate at 50% of execution
+    int checkAt = 5;
+    msg.set_cmdline(fmt::format("{} {}", checkAt, numLoops));
 
     // Call the functions
     sch.callFunctions(req);
 
+    // Sleep for a while to let the scheduler schedule the MPI calls, and then
+    // update the local slots so that a migration opportunity appears
+    SLEEP_MS(500);
```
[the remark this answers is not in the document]
csegarragonz: The request we are sending in the `sch.callFunctions` line is a request for an MPI application. Thus it is just one message, with the field `msg.mpiworldsize()` set to the `worldSize`.

Once the request is scheduled and has started executing, we will call `MpiWorld::create` that will then issue a _different_ call to `sch.CallFunctions` with `worldSize - 1` functions.

The way we make a migration opportunity appear in this test is by manually modifying the resources available in the main host by calling `sch.setThisHostResources`. On a high level, we do:

```
# Set low resource availability

# Call function with one function (will internally call the method another time)
sch.callFunctions

# Set high resource availability so that a migration opportunity appears
```

It is important than both the explicit (and the implicit) call to `sch.callFunctions` see the low resource availability. Hence the sleep.

Shillaker (approved):
Once https://github.com/faasm/faabric/pull/264 is merged, this is good to go.

PR #648, opened 2022-05-16
Added VSCode setup docs

mmathys:
cc @csegarragonz


csegarragonz (changes requested):
Looks good! Glad that you got it to work. I left some comments.

Also, we in general (even though we don't have a way to enforce it) try to stick to 80-char lines for Markdown when possible. Please try to reformat the text to follow this.

mmathys:
ok, I see.

PR #650, opened 2022-05-25
Latest WAMR

### LINKED ISSUE #635: Errors in WAMR filesystem support
Two tests are currently failing, both related to the WAMR filesystem implementation:

- `Test fread`: https://github.com/faasm/faasm/blob/main/tests/test/faaslet/test_filesystem.cpp
- `Test accessing shared files`: https://github.com/faasm/faasm/blob/main/tests/test/faaslet/test_shared_files.cpp 

These are perhaps related, perhaps not related, but both failures seem to be sporadic.

csegarragonz:
In this PR I bump the WAMR fork to the latest commit as of [today](3fd763a95c0493a0767e0aa54a77598e30a1d661), and cherry pick the codegen fixes from #642. This changes fix the broken tests in #635.

I also cleanup our [diff](https://github.com/bytecodealliance/wasm-micro-runtime/compare/main...faasm:faasm) with the upstream WAMR, minimising the changes to only those we need.

I have had to comment out one of the additions in #642, see [here](core/iwasm/compilation/aot_llvm.c). I will get on with the memory growth/shrinking in a separate PR.

Closes #635

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -131,7 +131,9 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        sanitiser: [None, Address, Thread, Undefined, Leak]
+        # 08/06/2022 - Disable Thread sanitiser build as it exhausts all the
+        # memory available in GHA runners
+        sanitiser: [None, Address, Undefined, Leak]
     env:
       CGROUP_MODE: off
       HOST_TYPE: ci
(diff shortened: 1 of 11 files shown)


Shillaker (approved):

csegarragonz:
@Shillaker it looks as though we may be running out of RAM for the TSAN build :/ 

[The available RAM for Linux GH hosted runners is 7 GB](https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources). I have tried increasing the swap space, but we are out of luck running inside a container, and I would be surprised if we could do that in a GitHub hosted runner anyway.

Let's discuss offline what's the best way to fix this.

PR #656, opened 2022-06-20
Push WASM memory management to `WasmModule`

csegarragonz:
In this PR I introduce support for `mmap/munmap` in WAMR. In the process, I push most of the memory management implementation from `WAVMWasmModule` to `WasmModule` to avoid duplication between WAMR and WAVM.

The implementation is still quite generic, and we can get away with few calls to the actual WASM runtime.

### DIFF
--- a/src/enclave/outside/ocalls.cpp
+++ b/src/enclave/outside/ocalls.cpp
@@ -75,18 +75,13 @@ extern "C"
     int32_t ocallSbrk(int32_t increment)
     {
         SPDLOG_TRACE("S - __sbrk - {}", increment);
-        wasm::WasmModule* module = wasm::getExecutingModule();
-        uint32_t oldBrk = module->getCurrentBrk();
-
-        if (increment == 0) {
-            return oldBrk;
-        } else if (increment < 0) {
-            module->shrinkMemory(-1 * increment);
-            return oldBrk;
-        } else {
-            return module->growMemory(increment);
-        }
+        SPDLOG_WARN("SGX-WAMR sbrk does not allocate more memory");
+        return 0;
     }
+
+    // ---------------------------------------
+    // Logging
+    // ---------------------------------------
 
     void ocallLogDebug(const char* msg) { SPDLOG_DEBUG("[enclave] {}", msg); }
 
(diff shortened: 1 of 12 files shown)


src/wasm/WasmModule.cpp
```diff
@@ -674,17 +674,125 @@ void WasmModule::writeWasmEnvToMemory(uint32_t envPointers, uint32_t envBuffer)
 
 uint32_t WasmModule::growMemory(size_t nBytes)
 {
-    throw std::runtime_error("growMemory not implemented");
+    if (nBytes == 0) {
```
csegarragonz, line 677: All the implementation in this file is copied from `WAVMWasmModule` with the necessary changes to make it runtime-agnostic.

src/wavm/WAVMWasmModule.cpp
```diff
@@ -1008,198 +1008,56 @@ U32 WAVMWasmModule::mmapFile(U32 fd, size_t length)
     return wasmPtr;
 }
 
-U32 WAVMWasmModule::growMemory(size_t nBytes)
+bool WAVMWasmModule::doGrowMemory(uint32_t pageChange)
```
csegarragonz, line 1011: This diff is very hard to read, but I have left the sense-checks that we can only do in WAVM. You can see the new code [here](https://github.com/faasm/faasm/blob/wasm-memory/src/wavm/WAVMWasmModule.cpp#L1011-L1059).

src/enclave/outside/ocalls.cpp
```diff
@@ -75,19 +75,14 @@ extern "C"
...
-
-        if (increment == 0) {
-            return oldBrk;
-        } else if (increment < 0) {
-            module->shrinkMemory(-1 * increment);
-            return oldBrk;
-        } else {
-            return module->growMemory(increment);
-        }
+        SPDLOG_WARN("SGX-WAMR sbrk does not allocate more memory");
```
csegarragonz, line 78: We don't really want to be growing memory outside the enclave. Memory management will have to be implemented inside the enclave.

Shillaker: Should this be an error then?

Shillaker (approved):
Nice, LGTM!

src/enclave/outside/ocalls.cpp
csegarragonz: We still call this method from SGX code, so even if we do nothing we must not error out in order for functions to finish.

PR #659, opened 2022-06-20
Re-introduce TSAN and configure ASAN

csegarragonz:
ASAN and TSAN builds of the tests are consuming upwards of 15 and 20 GB of memory each. This is bricking the Github-hosted runners.

In this PR, we modify the flags we use in these runs to reduce memory consumption. These tweaks also allow us to re-introduce TSAN. We also re-introduce some changes that fix some leaks.

Also for future reference, here are the [TSAN runtime flags](https://github.com/google/sanitizers/wiki/ThreadSanitizerFlags) and the [ASAN runtime flags](https://github.com/google/sanitizers/wiki/AddressSanitizerFlags).


.github/workflows/tests.yml
```diff
@@ -149,15 +149,11 @@ jobs:
     services:
       redis:
         image: faasm/redis:0.8.12
-        ports:
```
csegarragonz, line 152: GHA service containers and the job container are connected with a docker-specific network, so there is [no need to publish ports](https://docs.github.com/en/actions/using-containerized-services/about-service-containers#running-jobs-in-a-container):
> You don't need to configure any ports for service containers. By default, all containers that are part of the same Docker network expose all ports to each other, and no ports are exposed outside of the Docker network.

aktech:
2 cents on self-hosting runners:

I built a service for running self-hosted GitHub Action runners on cloud with 0 maintenance and free for open source: https://cirun.io/

- It's on-demand, which means you only pay to your cloud provider for the time you're using it. It also supports arm based runners.
- All you need is a simple configuration .cirun.yml file, here is a [demo](https://www.youtube.com/watch?v=lDx0YVTn6yg).
- Also available on GitHub Marketplace as well: https://github.com/marketplace/cirun-io

csegarragonz:
@eigenraven thanks for the tips! For ASAN, as you say, I think there might be another problem. For TSAN, changing the `history_size` to 0 does not help. I am a bit puzzled by that, as I was expecting a big reduce in memory consumption. However, it seems that TSAN's runtime flags need to be provided as a [space-spearated (and not colon-separated)](https://github.com/google/sanitizers/wiki/ThreadSanitizerFlags#runtime-flags) list. So we may have been using the default value of 2 all this time.

eigenraven:
@csegarragonz looking at the current failed run, there seems to be 1) a lot of useless tls access logging, 2) a lot of warnings about unreclaimed memory <https://github.com/faasm/faasm/runs/7000258660?check_suite_focus=true#step%3A23%3A5773=> - this looks like faasm logs rather than tsan logs

eigenraven:
Looks like this check is failing repeatedly <https://github.com/faasm/faasm/blob/main/src/wasm/WasmModule.cpp#L826-L833=>. I remember having to modify the brk mechanism locally because it didn't work properly when the wasm module allocated memory itself via `memory.grow`, and led to a lot of memory leaks if that function was called, maybe that's also what's happening now?

docker-compose.yml
```diff
@@ -129,9 +129,9 @@ services:
       - REDIS_QUEUE_HOST=redis-queue
       - REDIS_STATE_HOST=redis-state
       - LD_LIBRARY_PATH=/build/faasm/third-party/lib:/usr/local/lib
-      - ASAN_OPTIONS=halt_on_error=1
+      - ASAN_OPTIONS=halt_on_error=1:quqrantine_size_mb=16
```
eigenraven, line 132:
```suggestion
      - ASAN_OPTIONS=halt_on_error=1:quarantine_size_mb=16
```

docker-compose.yml
```diff
@@ -129,9 +129,9 @@ services:
...
       - LSAN_OPTIONS=suppressions=/usr/local/code/faasm/leak-sanitizer-ignorelist.txt
-      - TSAN_OPTIONS=halt_on_error=1:suppressions=/usr/local/code/faasm/thread-sanitizer-ignorelist.txt:history_size=7:second_deadlock_stack=1
+      - TSAN_OPTIONS="halt_on_error=1 suppressions=/usr/local/code/faasm/thread-sanitizer-ignorelist.txt history_size=0 flish_memory_ms=5000"
```
eigenraven, line 134:
```suggestion
      - TSAN_OPTIONS="halt_on_error=1 suppressions=/usr/local/code/faasm/thread-sanitizer-ignorelist.txt history_size=0 flush_memory_ms=5000"
```

csegarragonz:
@eigenraven thanks for pointing out the typos. FYI, the `flush_second_ms` really made the difference. Is there anything else that u think needs amending?

eigenraven (approved):
LGTM, out of curiosity what are the network namespaces data races? All functions there seem to be properly guarded by mutexes so it's a bit unexpected

csegarragonz:
Good question. I removed it and the tests pass as well. I can't remember why I included it in the first place...

PR #661, opened 2022-06-22
Remove Knative

csegarragonz:
In this PR we remove our integration with Knative. We were not using any of Knative's features, and it was making the deployment more complex, slow, and verbose.

In contrast, now we deploy Faasm on a bare k8s cluster. This simplifies the installation and deployment scripts. In summary, deploying/removing Faasm from an existing k8s cluster is a one-liner:

```bash
inv k8s.deploy --workers <num_workers> [--sgx]

inv k8s.delete
```

I think I have removed all dangling references to Knative in the org (but I am surely missing some). See related PRs: faasm/experiment-base#51, faasm/experiment-mpi#32, faasm/experiment-openmp#18


faasmcli/faasmcli/tasks/k8s.py
```diff
@@ -1,8 +1,8 @@
-from time import sleep
+from datetime import datetime
```
csegarragonz, line 1: Very messy diff. You may check the raw file [here](https://github.com/faasm/faasm/blob/remove-knative/faasmcli/faasmcli/tasks/k8s.py).

deploy/k8s/minio.yml.j2
```diff
@@ -11,7 +11,7 @@ metadata:
 spec:
   containers:
     - name: minio-main
-      image: faasm/minio:0.8.12
+      image: faasm/minio:{{ faasm_version }}
```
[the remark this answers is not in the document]
Shillaker: What other contents are we templating? I had thought I'd removed all templating from the repo as it makes it difficult to review changes, and generally adds more complexity. With this change we're now mixing SGX and non-SGX files again which I'd also quite like to avoid (just to keep the standard non-SGX setup as simple as possible).

faasmcli/faasmcli/tasks/k8s.py
```diff
@@ -142,84 +93,117 @@ def _deploy_faasm_services(sgx=False):
...
+                            "mountPath": "/var/run/aesmd",
+                        }
+                    ]
+                },
+                "extra_env_vars": [
+                    {"name": "SGX_AESM_ADDR", "value": "1"},
+                    {
+                        "name": "AZ_ATTESTATION_PROVIDER_URL",
+                        "value": "https://faasmattprov.eus2.attest.azure.net",
+                    },
```
Shillaker, line 162: If we were going to template this stuff, it might be cleaner to keep this in a YAML file, but wrap it with an `{% if sgx %}` to switch it on or off.

Shillaker (changes requested):
This looks good, other than the templating which I'm not sure about. Happy to discuss offline.

Shillaker (approved):
Yes, sorry for the delay. LGTM :+1:

PR #668, opened 2022-09-01
workon.sh fixes

### LINKED ISSUE #667: Fix workon.sh prefix
In [`workon.sh`](https://github.com/faasm/faasm/blob/bfb77185809a03a75113d99fe36b8c58112b7fa8/bin/workon.sh#L88), the `PS1` is updated the new virtual environment. If `workon.sh` is executed multiple times (e.g. when switching between environments), the old environment is kept in this variable, creating large string of virtual envs like this:

```bash
(faasm-exp) (faasm-exp) (faasm) (faasm-exp) (faasm) mmathys@koala9:~/experiment-base$
```

An easy fix could be to change this line to

```
export PS1="(faasm)"
```

mmathys:
closes #667

### DIFF
--- a/bin/workon.sh
+++ b/bin/workon.sh
@@ -13,7 +13,9 @@ if [[ -z "$FAASM_DOCKER" ]]; then
     if [ "$(ps -o comm= -p $$)" = "zsh" ]; then
         THIS_DIR="$( cd "$( dirname "${ZSH_ARGZERO}" )" >/dev/null 2>&1 && pwd )"
     fi
+    echo $THIS_DIR
     PROJ_ROOT="${THIS_DIR}/.."
+    #PROJ_ROOT="${THIS_DIR}/"
     VENV_PATH="${PROJ_ROOT}/venv-bm"
 
     # Normal terminal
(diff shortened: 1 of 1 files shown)


bin/workon.sh
```diff
@@ -37,6 +37,12 @@ pushd ${PROJ_ROOT} >> /dev/null
 # Virtualenv
 # ----------------------------
 
+# disable previous environment with `deactivate` if it exists
```
Shillaker, line 40: Bash file comments aren't markdown, so we don't need the backticks around `deactivate`. Comments should also start with a capital letter.

Shillaker (changes requested):

bin/workon.sh
```diff
@@ -13,7 +13,9 @@ if [[ -z "$FAASM_DOCKER" ]]; then
     if [ "$(ps -o comm= -p $$)" = "zsh" ]; then
         THIS_DIR="$( cd "$( dirname "${ZSH_ARGZERO}" )" >/dev/null 2>&1 && pwd )"
     fi
+    echo $THIS_DIR
     PROJ_ROOT="${THIS_DIR}/.."
+    #PROJ_ROOT="${THIS_DIR}/"
```
Shillaker, line 18: We shouldn't commit commented-out code. Is this needed?

Shillaker (changes requested):

Shillaker (approved):

PR #675, opened 2022-09-14
Prevent `inv cluster.stop` from removing CLI containers

mmathys:
With this change, the CLI containers are left running for development


mmathys:
@csegarragonz

csegarragonz (approved):

PR #680, opened 2022-10-05
Support to run ImageMagick

csegarragonz:
In this PR I include support to run ImageMagick in WAVM, and WAMR. The SGX PR will follow, as it is larger in size.

I also add a test that takes a sample image and flips it. I use Faasm's logo as test image.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -234,6 +234,11 @@ jobs:
     steps:
       - name: "Conan cache"
         uses: faasm/conan-cache-action@v1
+<<<<<<< HEAD
+=======
+        with:
+           directory: /usr/local/code/faasm
+>>>>>>> ef79d994 (gha: more workflow fixes)
       - name: "Ping redis"
         run: redis-cli -h redis ping
       - name: "Ping minio"
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -62,6 +62,7 @@ jobs:
           submodules: recursive
       - name: "Build C++ functions"
         run: ./bin/inv_wrapper.sh func.local
+        working-directory: ${{ github.workspace }}/clients/cpp
       - name: "Build libfake"
         run: ./bin/inv_wrapper.sh libfake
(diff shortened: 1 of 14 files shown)


include/wasm/WasmEnvironment.h
```diff
@@ -4,6 +4,20 @@
 #include <unordered_map>
 #include <vector>
 
+#define FAKE_NAME "faasm"
```
csegarragonz, line 7: Moving these definitions to a header where all runtimes can benefit from them.

src/wamr/filesystem.cpp
```diff
@@ -32,19 +33,47 @@ static uint32_t __wasi_fd_dup_wrapper(wasm_exec_env_t exec_env,
...
 static int32_t tempnam_wrapper(wasm_exec_env_t exec_env, int32_t a, int32_t b)
 {
     throw std::runtime_error("tempnam not implemented");
 }
 
 static NativeSymbol ns[] = {
-    REG_NATIVE_FUNC(__wasi_fd_dup, "(i*)i"),
-    REG_NATIVE_FUNC(dup, "(i)i"),
-    REG_NATIVE_FUNC(tempnam, "(ii)i"),
+    REG_NATIVE_FUNC(__wasi_fd_dup, "(i*)i"), REG_NATIVE_FUNC(dup, "(i)i"),
```
csegarragonz, line 74: Weird formatting by `clang-format` here.

.github/workflows/tests.yml
```diff
@@ -64,6 +64,12 @@ jobs:
       - name: "Build libfake"
         run: ./bin/inv_wrapper.sh libfake
         working-directory: ${{ github.workspace }}/clients/cpp
+      - name: "Update ImageMagick submodule"
+        run: git submodule update --init -f third-party/ImageMagick
+        working-directory: ${{ github.workspace }}/clients/cpp
```
Shillaker, line 69: Why do we need to update the ImageMagick submodule here? We're checking out the code with `submodules: true` which should do this implicitly.

csegarragonz:
This should be good after a rebase with #688

PR #681, opened 2022-10-07
SPDLOG-like logging in SGX

csegarragonz:


include/enclave/inside/ocalls.h
```diff
@@ -4,6 +4,25 @@
 #include <sgx.h>
 #include <sgx_defs.h>
 
+#include <stdio.h>
+
+// For code outside SGX enclaves we use spdlog macros for logging. We provide
+// similar macros for logging (only in debug mode) from inside the enclave. Note
+// that this logging is very costly as it does an ocall to print the message,
+// as a consequence we no-op it at compile time for release builds
```
Shillaker, line 12: It took me a while to work out that we no-op `ocallLogDebug`, and not the `SPDLOG_` macros here. Do we still call `snprintf` even in release builds? I would have expected:

```
#ifdef FAASM_SGX_DEBUG
#define SPDLOG_DEBUG_SGX(...) 
    size_t __bufferSize = 512;
    ...
#else
#define SPDLOG_DEBUG_SGX(...)
    // No-op
#endif
```

src/enclave/inside/EnclaveWasmModule.cpp
```diff
@@ -75,14 +75,12 @@ bool EnclaveWasmModule::callFunction(uint32_t argcIn, char** argvIn)
...
+        SPDLOG_DEBUG_SGX("Success calling WASM function");
     } else {
         std::string errorMessage(
           ((AOTModuleInstance*)moduleInstance)->cur_exception);
-        // TODO - better logging
-        std::string errorText =
-          "Caught WASM runtime exception: " + errorMessage;
-        ocallLogError(errorText.c_str());
+        SPDLOG_ERROR_SGX("Caught WASM runtime exception: %s",
+                         errorMessage.c_str());
```
Shillaker, line 83: I thought that spdlog formatting used `{}` not printf-style `%s`. Does this work?

Shillaker (changes requested):

PR #682, opened 2022-10-07
Bump WAMR's dependency to the most up-to-date version

csegarragonz:
WAMR is very much under development, and some of the patches we need to do are obsolete after some time. In addition, we periodically rebase our patches on the latest upstream head to get the newest features/fixes. As a consequence, I've squashed all our patches into a single commit in our [fork](https://github.com/faasm/wasm-micro-runtime). Now the [diff](https://github.com/faasm/wasm-micro-runtime/commit/caef0bc4db394b2d603066cc8b1a1cf5731e1b62) is very easy to track, and I will try to keep squashing the commits as I make progress.


Shillaker (approved):
Nice :+1:

PR #683, opened 2022-10-12
Asio+Beast-based endpoint server

eigenraven:
Update faabric to make use of the asio/beast-based endpoint server. Also fix a missing include in CGroup.cpp.


Shillaker (approved):

csegarragonz:
Are the `dist-tests` failures related to this PR?

eigenraven:
I have to check, it seems the error is with dns resolution, which could be because asio does it differently (or if dist-test-server crashes and doesn't print its logs out in the actions worker):
```
/usr/local/code/faasm/tests/dist/threads/test_pthreads.cpp:9: FAILED:
due to unexpected exception with message:
  Could not resolve host dist-test-server
```

eigenraven:
I can't replicate the crash locally, it seems to come from this call: `hostent* record = gethostbyname(hostname.c_str());` which is just the libc DNS resolver. I queued a rerun of dist-tests in case it was a weird failure on actions's side yesterday

eigenraven:
Found it, it's a asio segmentation fault because I didn't copy over the compiler definition to stop it from using RTTI. I'll push a fix in a minute

eigenraven:
I have pushed the fix that should work, but locally it doesn't for me - I'm really confused as to why, the crash is in a line that should be defined by a `#if !defined(...)` and checking the compiler logs that define is passed in the commandline. We'll see if this is just some weird bug with my local setup (though I did do a clean build) or there's something more going on

eigenraven:
I have to work on some other things right now, so I'll take a break from debugging this, if you want to try to reproduce this locally, try running `dist_test_server` from the cli container after building it in debug mode - it should crash with a SIGSEGV after a few seconds.

csegarragonz (approved):

PR #688, opened 2022-10-21
Fixes to CI/CD

### LINKED ISSUE #687: Fix deprecation warning in CI
Third-party actions that use `node:12` are marked as deprecated, so we should look for alternatives.

csegarragonz:
Part 3/3 of this PR. See faasm/faabric#279 for background.

Closes #687

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -59,7 +59,7 @@ jobs:
       - name: "Checkout code"
         uses: actions/checkout@v3
         with:
-          submodules: recursive
+          submodules: true
       - name: "Build C++ functions"
         run: ./bin/inv_wrapper.sh func.local
         working-directory: ${{ github.workspace }}/clients/cpp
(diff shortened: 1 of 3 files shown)


Shillaker (approved):

.github/workflows/tests.yml
```diff
@@ -110,10 +119,6 @@ jobs:
     steps:
       - name: "Conan cache"
         uses: faasm/conan-cache-action@v1
-        with:
-           directory: /usr/local/code/faasm
-      - name: "Build dev dependencies to be shared by all runs"
-        run: ./bin/inv_wrapper.sh dev.cmake --build Debug
```
eigenraven, line 116: Without running CMake this job becomes useless, the cache action only sets up the cache, while cmake populates it when it runs `conan install` during configuration.

csegarragonz: Thats a good point, thanks for bringing this up. The reason I removed it in the first place was because the `CMakeCache` would point to a different directory (so we needed the `--clean` and I thought that if we were to build with `--clean` then we might as well do it just once) but you are right.

PR #691, opened 2022-10-31
Remove `BOOST_NO_TYPEID` which snuck back in during one of the PR merges

eigenraven:
I added these 3 lines during experiments with #683, and later removed them with a rebase but they got merged in afterwards from the old version anyway. RTTI is fully enabled in the codebase now, so this is not needed and could lead to odd linker errors (odr violation) in the future.


csegarragonz (approved):
Gah! Sorry, my bad. Thanks for bringing this up!

PR #692, opened 2022-10-31
Pass `--cmdline` to `func_runner` and parse CL args

csegarragonz:
In this PR I modify the `func_runner` to optionally take a string to set as command line for the running function. I also take the oportunity to make the argc/argv parsing in `func_runner` more modular.

In summary, we can now do the following:

```bash
# Will print "hello echo"
inv run demo echo --data "hello echo"

# Will print 'hello' and 'world' as arguments
inv run demo argc_argv --cmdline "hello world"
```

As I am writing this examples, I'm thinking it would be a good idea to have integration tests for the FaasmCLI... :thinking:

### DIFF
--- a/faasmcli/faasmcli/tasks/run.py
+++ b/faasmcli/faasmcli/tasks/run.py
@@ -5,13 +5,15 @@ from faasmcli.util.shell import run_command
 
 
 @task(default=True)
-def run(ctx, user, function, data=None):
+def run(ctx, user, function, data=None, cmdline=None):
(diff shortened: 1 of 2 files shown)


Shillaker (approved):
Nice, good to add something more structured for this. 

Re. integration tests for Faasm CLI, I think we indirectly test it in the quick-start tests, but not sure if we do it elsewhere. Could be nice if it was a simple extra GHA step.

PR #697, opened 2022-11-02
Remove globbing from tests

csegarragonz:
This had been a long standing to-do that I address here. Remove the globbing for files and add them manually instead.

### DIFF
--- a/tests/dist/CMakeLists.txt
+++ b/tests/dist/CMakeLists.txt
@@ -1,4 +1,9 @@
-file(GLOB_RECURSE TEST_FILES ${CMAKE_CURRENT_LIST_DIR} test_*.cpp)
+# Add the different test directories. Each subdirectory will populate the
+# TEST_FILES variable, adding the source files for the tests
+add_subdirectory(chaining)
+add_subdirectory(mpi)
+add_subdirectory(state)
+add_subdirectory(threads)
 
 # Tests
 add_executable(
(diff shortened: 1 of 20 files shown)


Shillaker:
Can you remind me again what the problem with globbing is? This looks verbose and easy to miss tests by accident.

csegarragonz:
> Can you remind me again what the problem with globbing is? This looks verbose and easy to miss tests by accident.

@Shillaker see the `Note` in the [CMake docs](https://cmake.org/cmake/help/latest/command/file.html#filesystem). In my experience, the `Note` translates to having to remember to re-run `CMake` for the globbed file list to be re-generated, otherwise CMake has no way of knowing what files have changes. This is counter-intuitive, as it defeats a bit the purpose of globbing. On the other hand, adding the test names is more verbose, but once you get used to it you don't forget to add new files, and don't need to explicitly run `CMake` again.

Shillaker:
Hmm, I'm still worried that somebody will accidentally delete a test file or forget to add one. Is it possible to have a catch-all glob at the end as well or is that really gross?

csegarragonz:
Why is this a problem? AFAICT we list sources by name everywhere else in the the repo, why should tests be any different?

Shillaker:
If we leave sources out in the rest of the codebase, something will fail to compile (unless the code in that file isn't used anywhere, in which case it should be deleted). Tests are at the end of the include chain, i.e. they are not included by any other files implicitly or explicitly, so it's impossible to tell whether all of them are being run.

Shillaker (approved):

PR #698, opened 2022-11-03
Modify the correct command docker-comepose  

william-lbn:


Shillaker:
Hi @william-lbn, `docker compose` is correct, `docker-compose` is now deprecated. 

You need to update your Docker installation: https://docs.docker.com/engine/reference/commandline/compose/.

william-lbn:
> Hi @william-lbn, `docker compose` is correct, `docker-compose` is now deprecated.
> 
> You need to update your Docker installation: https://docs.docker.com/engine/reference/commandline/compose/.

thanks

PR #702, opened 2022-11-05
TFLite testcases

FlowDrucka64:


csegarragonz:
As discussed, this should now live in either [`faasm/examples`](https://github.com/faasm/examples) or [`faasm/experiment-tensorflow`](https://github.com/faasm/experiment-tensorflow).

PR #707, opened 2022-11-15
Replace VLA usage with heap-allocated buffers

eigenraven:
Variable length arrays are a compiler extension and not part of the base c++ language, they cause dynamic stack allocations which can silently overflow the stack, causing hard-to-debug errors at runtime which I have ran into a couple of times with large network messages.

The corresponding faabric pr <https://github.com/faasm/faabric/pull/284> should be merged first.


csegarragonz (approved):
LGTM. Good to be merged in after the faabric pr is merged in, and the submodule updated to point to faabric's `main`.

PR #719, opened 2023-02-08
Latest `wasi-libc`

csegarragonz:


src/wasm/WasmModule.cpp
```diff
@@ -67,121 +67,120 @@ WasmModule::WasmModule(int threadPoolSizeIn)
 
 WasmModule::~WasmModule() {}
 
-void WasmModule::flush() {}
```
csegarragonz, line 70: This diff is very convoluted and hard to read.

The reason is that I wanted to conditionally compile the memory-management-related functions (similarly to what I do in `include/wasm/WasmModule.h`). However, I realised that the function definitions where not groupped in the same way that we declare them in the header file, so I re-arrange them to be so.

If preferred, I can just surgically conditionally compile the functions across this file and undo the re-ordering.

I am also aware that defining functions in the order we declare them is not something we enforce, but I still think is a nice-to-have.

Shillaker (approved):

PR #721, opened 2023-02-13
Task to bump tracked submodule version and check submodule tracked branch

csegarragonz:
We want to make sure that the value of `./clients/[python,cpp]/VERSION` is the same than the one we use (i) in the gha file, and (ii) in `docker compose` (and transitively in the `quick-start` tests).

This PR adds:
* A function to bump the tracked version of the cpp or the python submodule.
* A check in GHA's that makes sure that the aforementioned versions are in sync.

I include the new check in the old `formatting` job, that I now re-factor to `checks`.

Whilst I am at it, I also include a check that makes sure that all submodules point to the `main` branch. On the down side, this means that a PR spanning multiple repos (as it happens often) will never be fully green until all submodules have been merged in. On the up side, this will prevent the very frequent issue of forgetting to update the branch after merge, which is quite non-intuitive.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -19,7 +19,7 @@ jobs:
       - name: Cancel Workflow Action
         uses: styfle/cancel-workflow-action@0.11.0
 
-  formatting:
+  checks:
     if: github.event.pull_request.draft == false
     runs-on: ubuntu-latest
     container:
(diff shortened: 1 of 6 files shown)


faasmcli/faasmcli/tasks/git.py
```diff
@@ -87,38 +88,63 @@ def _create_tag(tag_name, force=False):
...
     """
     Increase the version (defaults to bumping a single minor version)
     """
-    old_ver = get_faasm_version()
-
-    if ver:
-        new_ver = ver
+    bump_faasm_ver = (not python) and (not cpp)
+    if bump_faasm_ver:
+        old_ver = get_version()
```
csegarragonz, line 97: Faasm's bumping logic is the same, just indented.

.github/workflows/tests.yml
```diff
@@ -54,7 +65,7 @@ jobs:
     if: github.event.pull_request.draft == false
     runs-on: ubuntu-latest
     container:
-      image: faasm/cpp-sysroot:0.2.0
+      image: faasm/cpp-sysroot:0.2.1
```
csegarragonz, line 68: The motivation for this PR is that I realised that these versions were out of sync, and I have been forgetting often lately :facepalm:

.github/workflows/tests.yml
```diff
@@ -32,10 +32,21 @@ jobs:
...
-        run: git config --global --add safe.directory "$GITHUB_WORKSPACE"
+        run: |
+          git config --global --add safe.directory "$GITHUB_WORKSPACE"
+          git config --global --add safe.directory "$GITHUB_WORKSPACE/faabric"
+          git config --global --add safe.directory "$GITHUB_WORKSPACE/cpp"
+          git config --global --add safe.directory "$GITHUB_WORKSPACE/python"
       # Formatting checks
       - name: "Code formatting check"
-        run: ./bin/inv_wrapper.sh format-code
+        run: ./bin/inv_wrapper.sh format-code --check
```
csegarragonz, line 42: We were missing the `--check` flag, so formatting errors were slipping through....

PR #724, opened 2023-02-23
Fix release body generation

csegarragonz:
For some reason the release generating script using a third-party docker image was not working. My guess is that it was trying to index all the files in all the submodules, what was taking a very very long time.

Given that our requirements for release bodies are rather low, I implement a python function to generate the exact same release body.

Once the `wasi-libc`-related PRs are merged in, I will publish a new release.


Shillaker (approved):
Feels like there should be a library out there that does this off-the-shelf, but this works :+1:

PR #726, opened 2023-02-27
GHA: Azure Integration Tests

csegarragonz:
In this PR I add two workflows that run some tests on a production environment on Azure.

The first workflow runs the `demo/hello` function on a managed Kuberenetes cluster in AKS. It is only triggered on `workflow_dispatch`, which means that we need to navigate to the "Actions" tab, click on the workflow name (`Azure Integration Tests`), and click on the "Run workflow" button. The workflow could be improved in several ways:
* Run more complex functions (maybe some of the examples?).
* Be triggered on every tag push to the `main` branch. (The risk with this one is knowing when the images for the new tag will be ready to be pulled).
* Be triggered every push to `main` branch (even though noticeable changes will only happen once we release a new tag).

The second workflow runs the tests on an SGX-enabled Azure VM in hardware mode. It is triggered on `workflow_dispatch` and when a commit is pushed to the `main` branch (i.e. when a PR is merged in). This trigger behaviour aims to strike a compromise between not too often triggering yet often enough to catch regressions. Note that this action also provisions a fresh VM and installs Faasm from scratch, what also has helped flushing out installation issues.

To make it easier to access our Azure subscription without poluting the organisation's secrets, these workflows run on a `self-hosted` runner where we have already ran `az login`.

NB: Some of the SGX tests in Hardware mode are actually failing. I implement a plethora of fixes in #686, so rather than cherry-picking changes to make the tests pass here, I'd rather fix all SGX issues in #686.

### DIFF
--- a/tests/test/faaslet/test_env.cpp
+++ b/tests/test/faaslet/test_env.cpp
@@ -90,11 +90,11 @@ TEST_CASE_METHOD(MultiRuntimeFunctionExecTestFixture,
 
     SECTION("WAMR") { execWamrFunction(msg); }
 
-/* 04/03/2023 - This test is failing in hardware mode
-#ifndef FAASM_SGX_DISABLED_MODE
-    SECTION("SGX") { execSgxFunction(msg); }
-#endif
-*/
+    /* 04/03/2023 - This test is failing in hardware mode
+    #ifndef FAASM_SGX_DISABLED_MODE
+        SECTION("SGX") { execSgxFunction(msg); }
+    #endif
+    */
 }
 
 TEST_CASE_METHOD(MultiRuntimeFunctionExecTestFixture,


.env
```diff
@@ -1,6 +1,6 @@
-FAASM_VERSION=0.9.2
-FAASM_CLI_IMAGE=faasm/cli:0.9.2
-FAASM_WORKER_IMAGE=faasm/worker:0.9.2
+FAASM_VERSION=0.9.3
```
csegarragonz, line 1: I tag a new release to fix the errors in k8s files that the integration tests uncovered.

.github/workflows/azure.yml
```diff
@@ -11,8 +11,51 @@ defaults:
     shell: bash
```
csegarragonz, line 11: The reason this file appears as modified, and not as new, is because in order to test a workflow that is triggered on `workflow_dispatch`, the workflow needs to already be registered on the default branch.

Thus, I merged in two very simple PRs (#725, #727) to add the skeleton of the files.

.github/workflows/azure.yml
```diff
@@ -11,8 +11,54 @@ defaults:
...
+        run: ./bin/inv_wrapper.sh k8s.deploy --workers=4 --sgx ${{ matrix.sgx }}
+        working-directory: ${{ github.workspace }}/faasm
+      - name: "Build, upload and run a simple CPP function"
+        run: docker compose -f docker-compose-k8s.yml run -T cpp-cli ./bin/inv_wrapper.sh func demo hello func.upload demo hello func.invoke demo hello
+        working-directory: ${{ github.workspace }}/faasm
+      - name: "Delete cluster"
+        if: always()
+        run: ./bin/inv_wrapper.sh cluster.delete --name ${{ env.CLUSTER_NAME }}
+        working-directory: ${{ github.workspace }}/experiment-base
+      - name: "Chown all files"
```
csegarragonz, line 62: This is necessary to ensure that consecutive runs on the same runner can clean the state (i.e. remove all files). Otherwise, some docker-generated files may be `root`-owned.

.github/workflows/sgx_hw.yml
```diff
@@ -18,6 +18,79 @@ defaults:
...
+          ./bin/inv_wrapper.sh vm.run-command \
+            --name ${{ env.VM_NAME }} \
+            --path "code/faasm" \
+            --cmd "FAASM_BUILD_MOUNT=/build/faasm FAASM_LOCAL_MOUNT=/usr/local/faasm FAASM_CLI_IMAGE=faasm/cli-sgx:${{ env.FAASM_VERSION }} docker compose up --no-recreate -d faasm-cli cpp python"
+      - name: "Cross-compile CPP functions"
+        run: |
+          ./bin/inv_wrapper.sh vm.run-command \
+            --name ${{ env.VM_NAME }} \
+            --path "code/faasm" \
+            --cmd "sudo rm -rf ./clients/cpp/venv && docker compose exec cpp ./bin/inv_wrapper.sh func.local libfake"
```
csegarragonz, line 69: We remove the `./venv`'s to force their re-creation with the first call to `./bin/inv_wrapper.sh`, and to avoid possible inconsitencies between the new commits and the docker images.

Shillaker (approved):
This looks good, although these are some quite heavyweight tests to run on each main merge, especially for those unrelated to SGX/Azure-like functionality. 

Happy to merge as-is, but perhaps we could consider:

1. Run each night or once a week. 
2. Add a tag to a commit message to say that it should run the Azure integration tests (and by default switch them off)

PR #728, opened 2023-03-04
Fix race condition when running `./bin/cli.sh`

csegarragonz:
This one was hard to catch, and probably the cause of many `venv`-related problems.

In essence, `./bin/cli.sh` first makes sure that the cluster is up and running (`docker compose up -d --no-recreate`), and then `exec`s into one of the CLI containers.

In a fresh environment, the call to `docker compose up -d --no-recreate` will start the containers _and_ initialise the python virtual environment (i.e. will call `./bin/create_venv.sh`). The latter takes a while to finish, and if we exec into the CLI before it finishes (which calls `source ./bin/workon.sh`), we break the installation.

This can be reproduced by (literally):

```bash
git clone https://github.com/faasm/faasm
cd faasm
git submodule update --init -f
./bin/refresh_local.sh
./bin/cli.sh faasm
bash: /usr/local/code/faasm/bin/../venv/bin/activate: No such file or directory
# inv -l does not work
```

The quick start tests circunvent this issue miraculously:
* When we run `docker compose up`, we do it on the `nginx` service (that does _not_ start the `cpp` and `python` services).
* Then, when we start either the `cpp` or the `python` services, we `docker compose run` _not_ `docker compose exec` (as the services had not been started before).
* This means that when we call `./bin/inv_wrapper.sh` the `./venv` directory is initialised from scratch correctly.

This issue was brought up thanks to #726.

### DIFF
--- a/bin/cli.sh
+++ b/bin/cli.sh
@@ -47,21 +47,26 @@ if [[ -z "$1" ]]; then
     exit 1
 elif [[ "$1" == "faasm" ]]; then
     CLI_CONTAINER="faasm-cli"
+    VENV_ROOT=${PROJ_ROOT}/venv
 elif [[ "$1" == "faasm-sgx-sim" ]]; then
     CLI_CONTAINER="faasm-cli"
+    VENV_ROOT=${PROJ_ROOT}/venv
     export FAASM_CLI_IMAGE=${FAASM_SGX_CLI_IMAGE:-faasm/cli-sgx-sim:$(cat ${PROJ_ROOT}/VERSION)}
     export FAASM_WORKER_IMAGE=faasm/worker-sgx-sim:$(cat ${PROJ_ROOT}/VERSION)
     export WASM_VM=sgx
 elif [[ "$1" == "faasm-sgx" ]]; then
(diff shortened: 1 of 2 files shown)


cmake/ExternalProjects.cmake
```diff
@@ -111,7 +111,7 @@ FetchContent_Declare(wavm_ext
 
 FetchContent_Declare(wamr_ext
     GIT_REPOSITORY "https://github.com/faasm/wasm-micro-runtime"
-    GIT_TAG "135672cab24d877db4ef6933c0ab150351384d51"
+    GIT_TAG "a31e5a4fa299c4f8384f40e157b0a928ad0bda1b"
```
csegarragonz, line 114: Unfortunately, in a PR where I am working on SGX (#686) I rebased the `faasm` branch to the latest WAMR commit, effectively removing the previous tag. It does not exist anymore.

As a future note to self, any other time I rebase the WAMR dependency I will do it in a temporary branch whilst the PR is under development.

Shillaker (approved):

PR #729, opened 2023-03-04
Patch LLVM

csegarragonz:
See faasm/cpp#110

In this PR I bump the cpp and python submodules (and their versions) and, given that I make changes to the way we build the LLVM sysroot, I also update the `cpython` dep in the base docker image.


cmake/ExternalProjects.cmake
```diff
@@ -111,7 +111,7 @@ FetchContent_Declare(wavm_ext
 
 FetchContent_Declare(wamr_ext
     GIT_REPOSITORY "https://github.com/faasm/wasm-micro-runtime"
-    GIT_TAG "135672cab24d877db4ef6933c0ab150351384d51"
+    GIT_TAG "a31e5a4fa299c4f8384f40e157b0a928ad0bda1b"
```
csegarragonz, line 114: See [this comment](https://github.com/faasm/faasm/pull/728/files#r1125522410) (if you have not seen it yet).

Shillaker (approved):

PR #730, opened 2023-03-15
Migrate from DockerHub to ACR

csegarragonz:


.github/workflows/tests.yml
```diff
@@ -23,7 +23,10 @@ jobs:
     if: github.event.pull_request.draft == false
     runs-on: ubuntu-latest
     container:
-      image: faasm/cli:0.9.4
+      image: faasm.azurecr.io/cli:0.9.5
+      credentials:
```
csegarragonz, line 27: The `faasm.azurecr.io` is configured with `--anonymous-pull-enabled` so that images are still readily available to be used by anyone.

That being said, and given that this is a configurable option, I also include the credentials for pulling.

PR #731, opened 2023-03-22
MPI support in WAMR

csegarragonz:
In this PR I implement MPI support for WAMR.

Recently, execution of large applications in WAVM (e.g. LAMMPS) has been a bit flaky. I have also spotted areas of WAVM's WASI support that are out of sync with the spec. This is increasingly worrying after we have rebased to the latest `wasi-libc`.

This PR does not mean we are dumping WAVM, but it will help cross-check the origin of certain runtime errors. Also, and this speaks very well to the current WAVM MPI implementation and faasm/faabric split, implementing this was around a day's worth of work.

Most of the code is boilerplate code or WAMR-specific pointer/offset checking/translation. As an implementation note for the future, some of the MPI calls use double-pointers. When registering native symbols in WAMR, one may use the `*` character to indicate that a parameter (`int32_t`) is in fact an offset to WASM memory, and the WAMR runtime converts it into a pointer type, and does the corresponding bound checks so that it is ready to be read-from/written-to. However, if we are dealing with a double pointer, the second pointer is still an offset, not a readily usable native pointer.

NB: To run LAMMPS in WAMR we still need a couple more pieces that will come in a subsequent PR.

### DIFF
--- a/src/wamr/funcs.cpp
+++ b/src/wamr/funcs.cpp
@@ -3,16 +3,77 @@
 #include <faabric/util/bytes.h>
 #include <faabric/util/logging.h>
 #include <faabric/util/macros.h>
-
+#include <wamr/WAMRWasmModule.h>
 #include <wamr/native.h>
 #include <wasm/WasmExecutionContext.h>
 #include <wasm/WasmModule.h>
 #include <wasm/chaining.h>
+
 #include <wasm_export.h>
 
 using namespace faabric::scheduler;
 
 namespace wasm {
+
+static std::shared_ptr<faabric::state::StateKeyValue> getStateKV(
+  int32_t* keyPtr,
+  size_t size)
+{
+    WAMRWasmModule* module = getExecutingWAMRModule();
+    module->validateNativePointer(keyPtr, sizeof(int32_t));
+
+    const faabric::Message* call = &ExecutorContext::get()->getMsg();
+    char* key = reinterpret_cast<char*>(keyPtr); // second
+
+    faabric::state::State& s = faabric::state::getGlobalState();
+    auto kv = s.getKV(call->user(), key, size);
+
+    return kv;
+}
+
+/**
+ * Await a chained function's completion
+ */
+static int32_t __faasm_await_call_wrapper(wasm_exec_env_t exec_env,
+                                          int32_t callId)
+{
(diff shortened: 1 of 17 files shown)


src/wamr/funcs.cpp
```diff
@@ -51,38 +112,13 @@ static void __faasm_write_output_wrapper(wasm_exec_env_t exec_env,
...
-    return result;
-}
-
 static NativeSymbol ns[] = {
-    REG_NATIVE_FUNC(__faasm_write_output, "($i)"),
-    REG_NATIVE_FUNC(__faasm_read_input, "($i)i"),
-    REG_NATIVE_FUNC(__faasm_chain_ptr, "(i$i)i"),
     REG_NATIVE_FUNC(__faasm_await_call, "(i)i"),
+    REG_NATIVE_FUNC(__faasm_chain_ptr, "(i$i)i"),
+    REG_NATIVE_FUNC(__faasm_pull_state, "(*i)"),
```
csegarragonz, line 118: I have had to add a couple state functions as well for the MPI tests to pass.

src/wamr/funcs.cpp
```diff
@@ -51,38 +112,13 @@ static void __faasm_write_output_wrapper(wasm_exec_env_t exec_env,
     call.set_outputdata(outBuff, outLen);
 }
 
-/**
- * Chain a function by function pointer
- */
-static int32_t __faasm_chain_ptr_wrapper(wasm_exec_env_t exec_env,
```
csegarragonz, line 57: These two methods were just moved somewhere else in the file to preserve alphabetical order.

src/wamr/timing.cpp
```diff
@@ -56,13 +57,56 @@ uint32_t wasi_clock_time_get(wasm_exec_env_t exec_env,
...
-                          int32_t* d)
+                          int32_t* subscriptionsPtr,
+                          int64_t* eventsPtr,
+                          int32_t nSubs,
+                          int32_t* resNEvents)
 {
     SPDLOG_DEBUG("S - poll_oneoff");
-    throw std::runtime_error("poll_oneoff not implemented");
+
+    WAMRWasmModule* module = getExecutingWAMRModule();
```
csegarragonz, line 67: We also needed this function for the MPI tests.

PR #734, opened 2023-03-28
WAMR upgrade

csegarragonz:
In this PR I bump the WAMR reference, further simplify our diff with upstream WAMR, and use the newest public API calls to invoke wasm functions.

Most importantly, as part of the diff simplification, I remove some places in the WAMR codebase where we were hardcoding the memory layout. This was to prevent WAMR from doing some optimisations to the WASM's module memory. Instead of hacking WAMR (harder to maintain), I move to using a workaround in our compilation toolchain. I agree that all Faasm-compiled WASM modules will have this workaround (even if they don't run in WAMR), but the overhead is minimal (just a very small function that isn't even called), and really helps maintaining alignment with upstream WAMR. See faasm/cpp#113

I also make a couple of fixes here and there that make LAMMPS work on WAMR (!).

To-Do before merge:
* [x] Run WAMR examples faasm/examples#13
* [x] Squash WAMR diff to one commit: https://github.com/faasm/wasm-micro-runtime/commit/5e9dc3c7eb33167389d99b7e5851dc55b5911d33

This PR makes LAMMPS work on WAMR including migration (!).


src/enclave/inside/EnclaveWasmModule.cpp
```diff
@@ -60,26 +60,31 @@ bool EnclaveWasmModule::loadWasm(void* wasmOpCodePtr, uint32_t wasmOpCodeSize)
...
-
-    prepareArgcArgv(argcIn, argvIn);
+    if (func == nullptr) {
+        ocallLogError("Did not find named WASM function");
+        throw std::runtime_error("Did not find named wasm function");
+    }
 
     // Set dummy argv to capture return value
     std::vector<uint32_t> argv = { 0 };
+    bool success = wasm_runtime_call_wasm(execEnv, func, 0, argv.data());
```
csegarragonz, line 80: The only reason I change this here is because `aot_create_exec_env_and_call_function` is not even exposed as an API function anymore, so this file would not compile.

src/runner/local_pool_runner.cpp
```diff
@@ -32,7 +28,7 @@ int doRunner(int argc, char* argv[])
     usleep(1000 * 500);
 
     for (const auto& m : req->messages()) {
-        faabric::Message result = sch.getFunctionResult(m.id(), 20000);
+        faabric::Message result = sch.getFunctionResult(m.id(), 20000 * 100);
```
csegarragonz, line 31: For the local pool runner it makes sense to have longer timeouts, as we may want to run some longer running functions for debug purposes.

As a reminder, the local pool runner is a `pool_runner` that can be invoked from the command line. For example, the following runs a LAMMPS simulation without requiring any changes to faasm (i.e. any new `lammps_runner` etc):

```bash
 WASM_VM=wamr inv run.pool lammps main --cmdline '-in faasm://lammps-data/in.controller.wall'
```

src/wamr/CMakeLists.txt
```diff
@@ -11,6 +11,7 @@ set(WAMR_BUILD_SPEC_TEST 0)
 add_definitions(-DWAMR_FAASM=1)
 
 # Set AOT mode and JIT for code generation
+set(WAMR_BUILD_INTERPRETER 1)
```
csegarragonz, line 14: This was set by default implicitly through setting `WAMR_BUILD_AOT`, but I prefer to make it explicit.

src/wamr/WAMRWasmModule.cpp
```diff
@@ -231,51 +253,18 @@ int WAMRWasmModule::executeWasmFunction(const std::string& funcName)
...
-    bool success = aot_create_exec_env_and_call_function(
-      reinterpret_cast<AOTModuleInstance*>(moduleInstance),
-      reinterpret_cast<AOTFunctionInstance*>(func),
-      0,
-      argv.data());
-
+    bool success = wasm_runtime_call_wasm(execEnv, func, 0, argv.data());
     uint32_t returnValue = argv[0];
 
-    // Check function result
```
csegarragonz, line 249: Luckily, WAMR's error checking has improved since we first wrote this code, so we don't need such cumbersome checks.

src/wamr/WAMRWasmModule.cpp
```diff
@@ -231,51 +253,18 @@ int WAMRWasmModule::executeWasmFunction(const std::string& funcName)
...
         throw std::runtime_error("Did not find named wasm function");
     }
-
     // Note, for some reason WAMR sets the return value in the argv array you
     // pass it, therefore we should provide a single integer argv even though
     // it's not actually used
     std::vector<uint32_t> argv = { 0 };
-
-    // Invoke the function
-    bool success = aot_create_exec_env_and_call_function(
```
csegarragonz, line 241: This function was never part of the public API, and I think its not even available anymore.

We use the public API function `wasm_runtime_call_wasm`.

src/wamr/mpi.cpp
```diff
@@ -61,37 +61,37 @@ class WamrMpiContextWrapper
...
+    // MPI_Request is in itself a faabric_request_t* so requestPtrPtr is a
+    // faabric_request_t**, which is a double wasm offset. Allocating memory
+    // for the second pointer from outside WASM is tricky and error-prone,
+    // so we use overwrite the pointer value with the actual pointed-to value
+    // (the pointer just points to one int, so we can do that).
+    void writeFaasmRequestId(int32_t* requestPtrPtr, int32_t requestId) const
     {
         module->validateNativePointer(requestPtrPtr, sizeof(MPI_Request));
         MPI_Request* requestPtr = reinterpret_cast<MPI_Request*>(requestPtrPtr);
-        faabric::util::unalignedWrite<faabric_request_t*>(
```
csegarragonz, line 71: This `unalignedWrite` was actually smashing some values in the WASM stack, what was making LAMMPS hang mid execution, and made me have a very entretaining week :rofl:

src/wasm/WasmModule.cpp
```diff
@@ -827,9 +827,11 @@ void WasmModule::unmapMemory(uint32_t offset, size_t nBytes)
         SPDLOG_TRACE("MEM - munmapping top of memory by {}", pageAligned);
         shrinkMemory(pageAligned);
     } else {
-        SPDLOG_WARN("MEM - unable to reclaim unmapped memory {} at {}",
-                    pageAligned,
-                    offset);
+        // TODO - this log statement should be a warning, but for some reason
```
csegarragonz, line 830: Unfortunately, I don't know what's going on with this, but I will look into it.

csegarragonz: This is now fixed as per faasm/cpp#113

PR #735, opened 2023-03-30
Clean FDs

csegarragonz:


src/wamr/WAMRWasmModule.cpp
```diff
@@ -125,6 +122,9 @@ void WAMRWasmModule::doBindToFunction(faabric::Message& msg, bool cache)
 
 void WAMRWasmModule::bindInternal(faabric::Message& msg)
 {
+    // Prepare the filesystem
```
csegarragonz, line 125: For WAMR we need to make sure that this gets callled on every `reset`.

PR #755, opened 2023-05-23
Use file loader without local cache in the upload server

csegarragonz:
The behaviour of the upload server should be that, irrespective of whether the `user/function` has been uploaded before, the WASM bytes and the machine code are re-generated and uploaded to S3.

To achieve this goal, at some point in the past we introduced the file loader without local cache. As a reminder, the default behaviour of the file loader is to cache files in the local filesystem.

However, we were only using the file loader without cache to upload the WASM bytes, but not to generate the machine code. A `MachineCodeGenerator` instance takes a `FileLoader` in its constructor which, before, was configured to use the local cache.

This means that, even though the WASM file was uploaded, the generated machine code (which was later on uploaded to S3) corresponded to a potentially old WASM file. This was causing issues and making the `quick-start` tests fail. (Why this was not caught in the GHA is a different story that I cover in #746).

In this PR I fix this issue by explicitly giving `getMachineCodeGenerator` a `FileLoader&` and add a regression test. I also update the `FunctionLoaderTestFixture` to run upload tests for different WASM VMs.

### DIFF
--- a/tests/test/upload/test_upload.cpp
+++ b/tests/test/upload/test_upload.cpp
@@ -222,6 +222,59 @@ TEST_CASE_METHOD(UploadTestFixture, "Test upload and download", "[upload]")
 }
 
 TEST_CASE_METHOD(UploadTestFixture,
+                 "Test uploading function always overwrites",
+                 "[upload]")
+{
+    std::string fileKey = "gamma/delta/function.wasm";
+    std::string objFileKey;
+    std::string objFileHashKey;
+
+    SECTION("WAVM")
+    {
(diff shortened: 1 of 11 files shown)


tests/test/upload/test_upload.cpp
```diff
@@ -221,6 +221,76 @@ TEST_CASE_METHOD(UploadTestFixture, "Test upload and download", "[upload]")
     }
 }
 
+TEST_CASE_METHOD(UploadTestFixture,
```
csegarragonz, line 224: This test fails against the current `main` branch

